Issue #10, opened 2014-12-13
Add ECMAScript 6 features


nzakas:
This is a rollup that will let us keep track of ES6 progress.

- [x] `let` declarations
- [x] `const` declarations
- [x] Regular expression `u` flag
- [x] Regular expression `y` flag
- [x] Binary literals
- [x] New octal literals
- [x] Arrow functions
- [x] Default parameters
- [x] Rest parameters
- [x] Spread operator
- [x] `super` references
- [x] Classes
- [x] `for-of` loops
- [x] Generators
- [x] Object literal property shorthand
- [x] Object literal method shorthand
- [x] Object literal computed properties
- [x] Template strings
- [x] Destructuring
- [x] Unicode code point escapes
- [x] Allow duplicate object literal properties (except `__proto__`)
- [x] Modules

FredKSchott:
I'm starting work on Unicode code point escapes

nzakas:
Generators and code point escapes have been merged in.

nzakas:
Next priority is template literals. @FredKSchott interested in taking that?

FredKSchott:
definitely, I'll tackle that next (away for the long weekend but I'll take a look next week)

They're supported in V8 now, right? Makes sense to do them next

nzakas:
Yup! And also in io.js now.

I'm planning on attacking arrow functions when I have some energy.

xjamundx:
Interested in taking a shot in default params, though I'm not sure how much work it will be/take I'll have...

FredKSchott:
Template strings are done! I'd be interested in taking on destructuring or rest parameters next. @nzakas thoughts?

xjamundx:
+1 for destructuring.

FredKSchott:
@xjamundx whoops, just saw your post above on taking rest params. Are you still planning on working on that?

nzakas:
Rest params are much easier. Destructuring is tricky. You can decide the level of challenge you want. :)

FredKSchott:
![why not both](http://i0.kym-cdn.com/photos/images/newsfeed/000/538/731/0fc.gif)

nzakas:
:+1:

trusktr:
:+1: for modules and classes!

AbrahamAlcaina:
+1 classes

FredKSchott:
[...]

PR #30, opened 2014-12-29
New: Implement generators (refs #10)

nzakas:
Generators support for function declarations, function expressions, and object literal methods.


FredKSchott:
+1 good stuff!

PR #31, opened 2015-01-03
New: Add support for duplicate object literal properties

FredKSchott:
- Adds support for duplicate object properties in ES 6 with the `duplicateObjectLiteralProperties` feature. 
- Continues to forbid duplicate object literals in strict mode otherwise
- Special rules for the `__proto__` property

This feature doesn't exist in esprima-fb yet, so we couldn't use it to help generate tests. Instead all tests were created by this branch and then verified by me. It looks good to me, but a few more pairs of eyes couldn't hurt.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -2445,12 +2445,13 @@ function parseObjectProperty() {
 
 function parseObjectInitialiser() {
     var marker = markerCreate(),
+        allowDuplicates = extra.ecmaFeatures.objectLiteralDuplicateProperties,
         properties = [],
         property,
         name,
         key,
         kind,
-        map = {},
+        kindMap = {},
         toString = String;
 
     expect("{");
@@ -2459,20 +2460,29 @@ function parseObjectInitialiser() {
         property = parseObjectProperty();
 
         if (!property.computed) {
+
             if (property.key.type === astNodeTypes.Identifier) {
                 name = property.key.name;
             } else {
                 name = toString(property.key.value);
             }
 
-            /*eslint-disable no-nested-ternary*/
-            kind = (property.kind === "init") ? PropertyKind.Data : (property.kind === "get") ? PropertyKind.Get : PropertyKind.Set;
-            /*eslint-enable no-nested-ternary*/
+            if (property.kind === "init") {
+                kind = PropertyKind.Data;
+            } else if (property.kind === "get") {
+                kind = PropertyKind.Get;
+            } else {
+                kind = PropertyKind.Set;
+            }
 
             key = "$" + name;
-            if (Object.prototype.hasOwnProperty.call(map, key)) {
-                if (map[key] === PropertyKind.Data) {
-                    if (strict && kind === PropertyKind.Data) {
+            if (Object.prototype.hasOwnProperty.call(kindMap, key)) {
+                if (kindMap[key] === PropertyKind.Data) {
+                    if (kind === PropertyKind.Data && name === "__proto__" && allowDuplicates) {
+                        // Duplicate '__proto__' literal properties are forbidden in ES 6
+                        throwErrorTolerant({}, Messages.DuplicatePrototypeProperty);
+                    } else if (kind === PropertyKind.Data && strict && !allowDuplicates) {
+                        // Duplicate literal properties are only forbidden in ES 5 strict mode
                         throwErrorTolerant({}, Messages.StrictDuplicateProperty);
                     } else if (kind !== PropertyKind.Data) {
                         throwErrorTolerant({}, Messages.AccessorDataProperty);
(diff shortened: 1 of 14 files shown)


lib/features.js
```diff
@@ -66,6 +66,9 @@ module.exports = {
     // enable parsing of shorthand object literal properties
     objectLiteralShorthandProperties: false,
 
+    // Allow duplicate object literal properties (except '__proto__')
+    duplicateObjectLiteralProperties: false,
+
```
nzakas: I think it would be better to name `objectLiteralDuplicateProperties` to match the other flags that apply to object literals.

espree.js
```diff
@@ -2459,20 +2460,29 @@ function parseObjectInitialiser() {
...
             if (property.key.type === astNodeTypes.Identifier) {
                 name = property.key.name;
             } else {
                 name = toString(property.key.value);
             }
 
-            /*eslint-disable no-nested-ternary*/
-            kind = (property.kind === "init") ? PropertyKind.Data : (property.kind === "get") ? PropertyKind.Get : PropertyKind.Set;
-            /*eslint-enable no-nested-ternary*/
+            if (property.kind === "init") {
```
nzakas: Nice!

lib/features.js
FredKSchott: makes sense, done.

tests/fixtures/ecma-features/duplicateObjectLiteralProperties/proto-property.src.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+var proto = {};
+
+var x = {
+	__proto__: proto,
```
nzakas: Can you add another test using the string "__proto__" twice? It should not throw an error in that case.

nzakas:
This looks really good. I'd like to see tests using strings as identifiers. For instance, you can have duplicate "__proto__" properties if they are strings. So I think if you duplicate your current tests and use string property names, that should give us good enough coverage.

Also, be sure to update the readme with the new option.

FredKSchott:
Agreed, tests should be added for different kinds of property names. But I'm not sure if the ES 6 spec matches the behavior you're describing. See: http://people.mozilla.org/~jorendorff/es6-draft.html#sec-object.prototype.__proto__

For ***"\_\_proto\_\_ Property Names in Object Initializers"*** it says:
> It is a Syntax Error if PropertyNameList of PropertyDefinitionList contains any duplicate entries for "\_\_proto\_\_" and at least two of those entries were obtained from productions of the form `PropertyName : AssignmentExpression` .

Where `PropertyName` can be a `LiteralPropertyName` defined as:
```
PropertyName[Yield,GeneratorParameter] :
  LiteralPropertyName
  [+GeneratorParameter] ComputedPropertyName
  [~GeneratorParameter] ComputedPropertyName[?Yield]

LiteralPropertyName :
  IdentifierName
  StringLiteral
  NumericLiteral
```

So judging by the spec, `__proto__` duplicates will result in a syntax error if duplicates are defined as IdentifierName, StringLiteral, or ComputedPropertyName.

nzakas:
Looking at this closer, we are both wrong. :) I had sent a question about this to the mailing list and misunderstood a piece:
https://esdiscuss.org/topic/question-regarding-duplicate-proto-properties

`["__proto__"] : foo` is the thing that is treated differently. You cannot set a prototype using a computed property. This is reflected in step 6 of https://people.mozilla.org/~jorendorff/es6-draft.html#sec-__proto__-property-names-in-object-initializers.

So really the special case is with computed properties, not strings.

FredKSchott:
Ah you're right, thats exactly what the numbered steps are explaining. That's surprisingly satisfying...

The spec definitions themselves (not including those number steps) still seem to suggest that all PropertyNames (including computed) are valid for the `__proto`` special case. Do you think that is worth bringing up on es-discuss?

nzakas:
I don't think so. B.3.1 is the place in the spec that fully defines this behavior, including the various exceptions. It's referred to in other sections as well. The most relevant is the runtime semantics of PropertyDefinitionEvaluation: https://people.mozilla.org/~jorendorff/es6-draft.html#sec-object-initializer-runtime-semantics-propertydefinitionevaluation

FredKSchott:
Gotcha, I'm still unfamiliar with this spec / reading language specifications in general :)

nzakas:
No problem. So I think tasks here are too as a couple of tests with string property names and add a test that has both computed and duplicate properties enabled to try `["__proto__"]` (see ecma-features-mix directory for examples.

FredKSchott:
More tests added, let me know if there are any important cases that I missed

nzakas:
Lgtm. Thanks!

FredKSchott:
Follow up: I was still confused by the wording of the spec, so I ended up composing my confusion a bit more clearly and posting it to es-discuss. Looks like we were right on the behavior and the confusion was caused by an error in the spec.

the thread in question, for the curious: https://esdiscuss.org/topic/question-regarding-duplicate-computed-proto-properties

nzakas:
Nice!

PR #37, opened 2015-01-20
New: Add support for default parameters (refs #10)

xjamundx:
Basically a simple port of ariya/esprima@0384add.

Destructuring isn't a feature yet, so we're not testing that inside of default params, though it's pretty important as well. Let me know if I missed anything.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -3827,12 +3827,14 @@ function parseFunctionSourceElements() {
 }
 
 function parseParams(firstRestricted) {
-    var param, params = [], token, stricted, paramSet, key, message;
+    var param, params = [], defaults = [], defaultCount = 0, token, stricted, paramSet, key, message, def,
+        allowDefaultParams = extra.ecmaFeatures.defaultParams;
     expect("(");
 
     if (!match(")")) {
         paramSet = {};
         while (index < length) {
+            def = null;
             token = lookahead;
             param = parseVariableIdentifier();
             key = "$" + token.value;
@@ -3857,6 +3859,15 @@ function parseParams(firstRestricted) {
                     message = Messages.StrictParamDupe;
                 }
             }
+            if (match("=")) {
+                if (!allowDefaultParams) {
+                    throwUnexpected(lookahead);
+                }
+                lex();
+                def = parseAssignmentExpression();
+                ++defaultCount;
+            }
+            defaults.push(def);
             params.push(param);
             paramSet[key] = true;
             if (match(")")) {
@@ -3868,9 +3879,14 @@ function parseParams(firstRestricted) {
 
     expect(")");
 
+    if (defaultCount === 0) {
+        defaults = [];
+    }
+
     return {
         params: params,
         stricted: stricted,
+        defaults: defaults,
         firstRestricted: firstRestricted,
         message: message
     };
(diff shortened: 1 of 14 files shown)


espree.js
```diff
@@ -3857,6 +3858,15 @@ function parseParams(firstRestricted) {
...
             }
+            if (match("=")) {
+                if (!allowDefaultParams) {
+                    throwUnexpected(lookahead);
+                }
+                lex();
+                def = parseAssignmentExpression();
+                ++defaultCount;
+            }
+            defaults.push(def);
```
FredKSchott: Is there a reason to always push `def` into `defaults`, even when a `=` isn't matched/found?

FredKSchott:
It would be good to add tests to cover functions with a mix of default and default-less params
(`function(a, b = 42, c)`)

nzakas:
Looks like a good start. Can you see what tests Esprima has for default parameters? I'd like to make sure we have the same ones.Check https://github.com/ariya/esprima/blob/harmony/test/harmonytest.js

espree.js
xjamundx: good call. may have mis-copied? will check.

nzakas: Also keep in mind that Esprima does have various bugs with ES6 features that haven't been fixed. Just because it's in Esprima doesn't mean it's correct.

xjamundx:
Hey @nzakas I copied them verbatim. The remaining ones all rely on destructuring or fat arrows.

nzakas:
Ah okay, nevermind then. :)

xjamundx:
Working on `function(a, b = 42, c))` is uncovering some minor issues :) Yay for more tests!

espree.js
xjamundx: Answer is yes it needs to be there, so that the `defaults` array can be filled with `null` values in the case that there isn't a default for one of the arguments. (See latest update)

xjamundx:
FYI for the tests I'm pulling my `.results.js` by pasting the source into http://esprima.org/demo/parse.html with both **Index-based range** and  **Line and column-based** checked.

nzakas:
There's an easier way to do that. You can use the test generator. See `tools/create-test.js` and `tools/create-test-example.js`. The script will automatically create the results file for you. :)

espree.js
FredKSchott: Gotcha, that makes sense. Then is the `defaultCount === 0` check below still necessary?

xjamundx: Checked and yes it is.

For example if you have `function x(a, b, c){}`

The `defaults` should equal `[]` instead of `[null, null, null]`.

The `null` are only expected as placeholders when there is a partially full list of default parameters.

FredKSchott: :+1: Sounds good, thanks for double-checking

xjamundx:
May be wrong but I think this is :+1:

nzakas:
Looks good. Can you add a test using a shorthand method? This should go in `ecma-features-mix`. See the examples already there.

xjamundx:
Okay, tried that and also used your `tools/create-test.js` thing which seemed to work well. :+1:

nzakas:
Looks good, thanks!

PR #38, opened 2015-01-21
New: Add support for template strings

FredKSchott:
Adds support for template strings, expressions, tagging, etc. (see #10)


FredKSchott:
The good engineer in me wants to clean up `scanTemplate()` to better leverage `scanStringLiteral()`, but I don't want to move our code too far away from esprima. @nzakas thoughts?

espree.js
```diff
@@ -921,6 +921,170 @@ function scanStringLiteral() {
...
+                ++index;
+                terminated = true;
+                break;
+            }
+            cooked += ch;
+        } else if (ch === "\\") {
+            ch = source[index++];
+            if (!ch || !syntax.isLineTerminator(ch.charCodeAt(0))) {
+                switch (ch) {
+                case "u":
```
nzakas: Can you indent the cases? I hate this style. :)

espree.js
```diff
@@ -2094,6 +2264,10 @@ function throwUnexpected(token) {
         throwError(token, Messages.UnexpectedToken, token.value);
     }
 
+    if (allowTemplateStrings && token.type === Token.Template) {
```
nzakas: `allowTemplateStrings` might not be necessary here.

espree.js
```diff
@@ -2634,6 +2847,8 @@ function parsePrimaryExpression() {
             expr = astNodeFactory.createLiteralFromSource(scanRegExp(), source);
         }
         peek();
+    } else if (allowTemplateStrings && type === Token.Template) {
```
nzakas: Ditto. Since you can only get a `Token.Template` token when template strings are enabled, you don't need to check both conditions. If the token is `Token.Template`, then you know template strings are enabled.

espree.js
```diff
@@ -2706,55 +2921,54 @@ function parseNewExpression() {
...
     state.allowIn = true;
     expr = matchKeyword("new") ? parseNewExpression() : parsePrimaryExpression();
     state.allowIn = previousAllowIn;
 
-    for (;;) {
-        if (match(".")) {
-            property = parseNonComputedMember();
-            expr = astNodeFactory.createMemberExpression(".", expr, property);
-        } else if (match("(")) {
+    while (match(".") || match("[") || match("(") || (allowTemplateStrings && lookahead.type === Token.Template)) {
```
nzakas: Same here.

espree.js
```diff
@@ -2706,55 +2921,54 @@ function parseNewExpression() {
...
+        allowTemplateStrings = extra.ecmaFeatures.templateStrings,
+        previousAllowIn = state.allowIn,
         marker = markerCreate();
 
-    previousAllowIn = state.allowIn;
     expr = matchKeyword("new") ? parseNewExpression() : parsePrimaryExpression();
     state.allowIn = previousAllowIn;
 
-    while (match(".") || match("[")) {
+    while (match(".") || match("[") || (allowTemplateStrings && lookahead.type === Token.Template)) {
```
nzakas: And here :)

lib/features.js
```diff
@@ -3,6 +3,7 @@
  *      settings.
  * @author Nicholas C. Zakas
  * @copyright 2014 Nicholas C. Zakas. All rights reserved.
+ * @copyright 2015 Fred K. Schott. All rights reserved.
```
nzakas: Newest copyright should be first. :)

nzakas:
If you think there's a way to reduce duplication, I'd say go for it. You have the tests covering you.

FredKSchott:
Comments addressed and duplication reduced, ready for review

espree.js
```diff
@@ -799,107 +800,120 @@ function scanNumericLiteral() {
...
+function scanEscapeSequence(ch) {
+    var code,
+        unescaped,
+        restore,
+        escapedCh,
+        octal = false;
+
+    // An escape sequence cannot be empty
+    if (!ch) {
+        throwError({}, Messages.UnexpectedToken, "ILLEGAL");
```
FredKSchott: The only change this refactoring made to the behavior itself was to throw an error if this case is reached. If you check back in the removed code, it looks like this case currently gets lost in the pile of if/elses and goes unchecked, when it should result in an error. 

Does this sound correct? A second pair of eyes couldn't hurt

nzakas: I think this case can't happen in the old code. The loop prevents reading past the end of the source text so `ch` is always valid.

FredKSchott: The old code did this:          
```
} else if (ch === "\\") {
  ch = source[index++];
  if (!ch || !syntax.isLineTerminator(ch.charCodeAt(0))) {
    // parse escape sequence
  } else {
    ++lineNumber;
     if (ch === "\r" && source[index] === "\n") {
      ++index;
    }
    lineStart = index;
  }
```

Since index keeps getting incremented, its hard to see what's being checked and what isn't. It looks like `!ch` was meant to be checked but instead it was just being treated like a line terminator.

nzakas: Hmmm, yeah, interesting. What you have is probably find regardless.

nzakas:
Looks good to me!

PR #45, opened 2015-01-31
Fix: Tokenization for template strings (fixes #44)

### LINKED ISSUE #44: Template string tokens are strange
The way that template string tokens are generated is pretty strange. This code:

```js
var foo = `hi${bar}`;
```

Is tokenized as:

```
[
    {
        "type": "Keyword",
        "value": "var"
    },
    {
        "type": "Identifier",
        "value": "foo"
    },
    {
        "type": "Punctuator",
        "value": "="
    },
    {
        "value": "`hi${"
    },
    {
        "type": "Identifier",
        "value": "bar"
    },
    {
        "type": "Punctuator",
        "value": "}"
    },
    {
        "type": "Punctuator",
        "value": ";"
    }
]
```

So the first part of the template string is represented by a token without a type, and then the rest are just regular tokens, completely losing the end of the template string.

nzakas:
Template strings really mess with tokenization because the tokens actually are different depending on the context before a `}` is encountered. Yuck. But the nice thing now is that this code:

```
`Hi ${foo} you.`;
```

Will produce:

1. `` "`Hi ${" `` (Template)
2. `"foo"` (identifier)
3. `` "} you.`" `` (Template)

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -936,8 +936,10 @@ function scanStringLiteral() {
 }
 
 /**
- * Scan a template string and return an ASTNode representation
- * @returns {ASTNode} The template string literal
+ * Scan a template string and return a token. This scans both the first and
+ * subsequent pieces of a template string and assumes that the first backtick
+ * or the closing } have already been scanned.
+ * @returns {Token} The template string token.
  * @private
  */
 function scanTemplate() {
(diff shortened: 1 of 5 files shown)


nzakas:
@FredKSchott just FYI. This was really buggy in Esprima, thought you'd like to see how I fixed it.

nzakas:
Worth noting: I'm pretty sure this is the way that Esprima intended this to work, it just never worked correctly.

@jeffmo

FredKSchott:
I completely missed how gross this was. Nice catch & fix!

Since `${}` is parsed out of the template string as an expression token, does it make sense to keep it in the template string value that espree returns? Is this only to match esprima, or is there a good reason to keep it regardless?

nzakas:
The `${}` isn't a separate token, it's part of the template, similar to how the backtick is a part of it or how quotes are a part of strings.

FredKSchott:
Gotcha, thanks for clarifying

PR #47, opened 2015-02-01
New: Allow return in global scope (fixes #46)

### LINKED ISSUE #46: Allow return in global scope
For Node.js compatibility, it would be good to add the option to use `return` in the global scope.

Working on this now.

nzakas:


FredKSchott:
Is this an ES6 feature? Any docs on it for reference? (global + return is not the most valuable search)

nzakas:
I don't want to introduce multiple ways to turn features off and on when the only real difference is semantics. JSX also isn't an ECMA standard, but it's in `ecmaFeatures` because I view that as saying "tell me how you want to parse ECMAScript".

PR #48, opened 2015-02-01
New: Allow super references in functions (refs #10)

nzakas:


nzakas:
FWIW, the option name is `superInFunctions` because opting-in to classes should also opt-in to using `super` within class methods. Having `super` in functions is separate because there's no dependency.

FredKSchott:
+1 lgtm

PR #61, opened 2015-02-17
Fix: Extend property method range and loc to include params (fixes #36)

### LINKED ISSUE #36: Object literal shorthand generator methods have incorrect range
Perhaps this is intentional, but the range given to object literal shorthand generator methods includes only the method body `{}`. Since they key is part of the parent property and overlapping, non-ancestral ranges might get confusing, I'm not sure that the `FunctionExpression`'s range should include all of `*foo(arg1) {}`, but it seems like the range should expand to at least `(arg1) {}`.

Related: eslint/eslint#1677

```js
//test.js
var inspect = require('util').inspect,
	espree = require('espree');

var source = [
	"({",
	"    *foo(arg1) {}",
	"})"
].join("\n");

var ast = espree.parse(source, {
	range: true,
	tokens: true,
	ecmaFeatures: {
		generators: true,
		objectLiteralShorthandMethods: true
	}
});

console.log(inspect(ast.body[0].expression.properties[0]));
console.log(inspect(ast.body[0].expression.properties[0].value.params[0]));
console.log(inspect(ast.tokens));
```

```sh
$  node test.js
{ type: 'Property',
  key: { type: 'Identifier', name: 'foo', range: [ 8, 11 ] },
  value:
   { type: 'FunctionExpression',
     id: null,
     params: [ [Object] ],
     defaults: [],
     body: { type: 'BlockStatement', body: [], range: [Object] },
     rest: null,
     generator: true,
     expression: false,
     range: [ 18, 20 ] },
  kind: 'init',
  method: true,
  shorthand: false,
  computed: false,
  range: [ 7, 20 ] }
{ type: 'Identifier', name: 'arg1', range: [ 12, 16 ] }
[ { type: 'Punctuator', value: '(', range: [ 0, 1 ] },
  { type: 'Punctuator', value: '{', range: [ 1, 2 ] },
  { type: 'Punctuator', value: '*', range: [ 7, 8 ] },
  { type: 'Identifier', value: 'foo', range: [ 8, 11 ] },
  { type: 'Punctuator', value: '(', range: [ 11, 12 ] },
  { type: 'Identifier', value: 'arg1', range: [ 12, 16 ] },
  { type: 'Punctuator', value: ')', range: [ 16, 17 ] },
  { type: 'Punctuator', value: '{', range: [ 18, 19 ] },
  { type: 'Punctuator', value: '}', range: [ 19, 20 ] },
  { type: 'Punctuator', value: '}', range: [ 21, 22 ] },
  { type: 'Punctuator', value: ')', range: [ 22, 23 ] } ]
```

*Edit: Add `arg1` to output.*

btmills:
This is jquery/esprima#1036 ported to Espree.

Previously, the range and loc for get, set, and shorthand methods inside object literal properties included only the method body, meaning a method's parameters were outside the range of their parent `FunctionExpression` node. This was happening because the marker for the `FunctionExpression` node wasn't being created until `parsePropertyFunction` was called, by which point the mothod's parameters had already been parsed. This PR solves the issue by creating the marker for the start of the `FunctionExpression` node as soon as it determines that the property is a get, set, or shorthand method, then passing the marker into `parsePropertyFunction`.

Though #36 only describes the specific case of generator shorthand methods, the same bug is present on get, set, and shorthand property methods, and this PR fixes all three cases.

The only functional change here is in `espree.js` itself; the other files are all updated tests.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -2375,8 +2375,7 @@ function parsePropertyFunction(options) {
         previousYieldAllowed = state.yieldAllowed,
         params = options.params || [],
         defaults = options.defaults || [],
-        body,
-        marker = markerCreate();
+        body;
 
     state.yieldAllowed = options.generator;
 
@@ -2393,7 +2392,7 @@ function parsePropertyFunction(options) {
     strict = previousStrict;
     state.yieldAllowed = previousYieldAllowed;
 
-    return markerApply(marker, astNodeFactory.createFunctionExpression(
+    return markerApply(options.marker, astNodeFactory.createFunctionExpression(
         null,
         params,
         defaults,
@@ -2406,6 +2405,7 @@ function parsePropertyFunction(options) {
 
 function parsePropertyMethodFunction(options) {
     var previousStrict = strict,
+        marker = markerCreate(),
         tmp,
         method;
 
@@ -2421,7 +2421,8 @@ function parsePropertyMethodFunction(options) {
         params: tmp.params,
         defaults: tmp.defaults,
         rest: tmp.rest,
-        generator: options ? options.generator : false
+        generator: options ? options.generator : false,
+        marker: marker
     });
 
     strict = previousStrict;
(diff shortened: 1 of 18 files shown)


nzakas:
Nice, thanks.

Issue #56, opened 2015-02-09
support destructured params w/defaults (shorthand syntax)


xjamundx:
While support for destructuring was added in #51 none of these cases work:
- `function a({x = 10}) {}` 
- `function a({x = 10, y: { z = 10 }}) {}`
- `({x = 10}) => x`
- `({x = 10, y: { z = 10 }}) => [x,y]`
- `[a=10] = 0`
- `var {a = 10} = {}`

Latest spec can be found here:
https://github.com/estree/estree/issues/13

xjamundx:
Will spend like 40 minutes trying to get this to work and see what I can come up with...

nzakas:
Cool. Just as an FYI, I'm looking at how we can potentially merge back into Esprima now that it's moving again. I won't stop you if you want to tackle this, just wanted to give to all the info before you spend more time.

xjamundx:
That's great news, we'll see how far along I get. It's good learning :)

xjamundx:
Related https://github.com/jquery/esprima/issues/1028

xjamundx:
Will try to take another stab at this sometime later this week...

PR #62, opened 2015-02-19
Fix: Destructuring defaults

xjamundx:
Refs #56 

- [x] not throwing errors when encountering destructured defaults
- [x] produce spec-compliant AST
- [x] all tests pass
- [x] more tests


xjamundx:
Okay, finally tracked down a spec I could use:

The acorn implementation of `var {x = 10} = x` looks like:

```js
{ type: 'Program',
  start: 0,
  end: 16,
  body: 
   [ { type: 'VariableDeclaration',
       start: 0,
       end: 16,
       declarations: 
        [ { type: 'VariableDeclarator',
            start: 4,
            end: 16,
            id: 
             { type: 'ObjectPattern',
               start: 4,
               end: 12,
               properties: 
                [ { type: 'Property',
                    start: 5,
                    end: 11,
                    method: false,
                    shorthand: true,
                    computed: false,
                    key: { type: 'Identifier', start: 5, end: 6, name: 'x' },
                    kind: 'init',
                    value: 
                     { type: 'AssignmentPattern',
                       start: 5,
                       end: 11,
                       operator: '=',
                       left: { type: 'Identifier', start: 5, end: 6, name: 'x' },
                       right: { type: 'Literal', start: 9, end: 11, value: 10, raw: '10' } } } ] },
            init: { type: 'Identifier', start: 15, end: 16, name: 'x' } } ],
       kind: 'var' } ] }
```

nzakas:
Nice!

xjamundx:
- 2 tests with minor column number issues.
- 3 failing because we're not handling the nested cases
- 1 more failing just because...

xjamundx:
5 failing due to missing `range` property, should be minor stuff now.

xjamundx:
Okay fixed that now just need to add a few more tests!

xjamundx:
Okay this should be :+1: 

Code is all mine, because nothing in Esprima to copy, so seeking a good review!

espree.js
```diff
@@ -2560,6 +2560,26 @@ function parseObjectProperty() {
             );
         }
 
+        // destructuring defaults
+        if (allowDestructuring && match("=")) {
+            lex();
+            var value = parseAssignmentExpression();
+            var prop = markerApply(marker, astNodeFactory.createAssignmentExpression("=", id, value));
+            prop.type = astNodeTypes.AssignmentPattern;
+            var shorthand = true; // TODO: figure out the actual value
```
xjamundx: kind of unclear why this only works when hard-coded to `true`. We added non-shorthand tests and they seem to pass as well....

espree.js
```diff
@@ -4356,7 +4384,7 @@ function parseParam(options) {
         if (rest) {
             throwErrorTolerant(lookahead, Messages.DefaultRestParameter);
         }
-        if (!allowDefaultParams) {
+        if (!(allowDefaultParams || allowDestructuring)) {
```
nzakas: This reads a bit funny, can you clean it up?

espree.js
nzakas: What happens when it's not hardcoded?

espree.js
xjamundx: `if (!allowDefaultParams || !allowDesctructuring)` has a slightly diff meaning and won't work.

how about:

```js
if (allowDefaultParams || allowDestructuring) {
	lex();
	def = parseAssignmentExpression();
	++options.defaultCount;
} else {
	throwUnexpected(lookahead);
}
```

xjamundx:
Okay give this one a lookover if you can. I figured out why shorthand was required in that block. Longform is a different code path and probably already worked :)

espree.js
nzakas: :+1:

nzakas:
Lgtm. Good to merge?

xjamundx:
Yeah I think it's :+1:

PR #67, opened 2015-02-28
New: Support ES6 classes (refs #10)

nzakas:
Three of the original Esprima tests are failing due to a strict mode violation. For some reason, my changes introduced a problem where things like this aren't flagged:

```js
function f() {"use strict"; x={ set s(eval){} }}
```

It should throw an error on `eval`, since that's now allowed in strict mode, but doesn't. The three test failures all have to do with this situation (same with getters and methods as well).

I've been staring at this for too long and can't figure out how to fix it. Any insights greatly appreciated. :)


espree.js
```diff
@@ -4565,14 +4688,149 @@ function parseYieldExpression() {
...
+
     if (lookahead.type === Token.Keyword) {
         switch (lookahead.value) {
             case "function":
                 return parseFunctionDeclaration();
+            case "class":
+                if (allowClasses) {
+                    return parseClassDeclaration();
+                }
+                break;
```
btmills: Reading through from the start trying to grok this. If classes are disabled, shouldn't this go to the `default` case as with block bindings below?

btmills: Ah, I see. The `default` case and `break`ing out of the `switch` both end up calling `parseStatement()`. Doesn't that make the `default` case redundant?

espree.js
```diff
@@ -4565,14 +4688,149 @@ function parseYieldExpression() {
...
+        computed = match("[");
+        marker = markerCreate();
+        key = parseObjectPropertyKey();
+
+        if (key.name === "static" && lookaheadPropertyName()) {
+            token = lookahead;
+            isStatic = true;
+            computed = match("[");
+            key = parseObjectPropertyKey();
+            console.dir(key);
```
btmills: Assuming this is useful for debugging. Noting here to make sure it gets removed before merging.

nzakas: Yikes, thought I had removed all of those.

espree.js
nzakas: Yeah, probably. I mostly just copy-pasted. :)

btmills:
[espree.js line 2613](https://github.com/eslint/espree/pull/67/files#diff-7a476a3203e745831344f37dd24a8a97R2613) in the `set` portion of `parseObjectProperty()` is `param = [ parseVariableIdentifier() ];`. Nowhere does it call `validateParam()`, which gets called by `parseParam()` in normal methods. Could that be the issue?

nzakas:
Ah could be! That method was completely refactored in Esprima, and I really didn't want to fight through that refactoring with this commit.

nzakas:
@btmills That was it! All tests are passing! We have classes!

Issue #72, opened 2015-03-05
Update module functionality


nzakas:
* [x] add test to validate that `sourceType="module"` effectible enables strict mode (e.g.: `WithStatement` is invalid in modules)
* [x] `ImportDeclaration` and `ExportDeclaration` cannot be used inside functions (missing tests and implementation)
* [x] Enable classes for export and export default declarations (all TODOs are in the code, very straight forward)

nzakas:
Looking at usage patterns, I think I'd also like to make a small change:

1. Add a `modules` feature flag that just enables import/export declarations and switches the parser into strict mode.
1. Make `sourceType = "module"` turn on every ES6 feature flag, including `modules`.

I think that gives us both parity with Esprima and the configurability some have asked for in ESLint.

caridy:
ha, I told you so!, lets roll with that, should be flexible enough. I wonder if the inverse will be possible. Using `sourceType="module"` with `class` disabled. This is sort of a requirement that we have at Yahoo, where we want to focus on zero-cost abstraction features only rather than all. e.g.: we want to use modules but we don't want to use `let`.

nzakas:
Yup, you were right.

I think in this model, you can use `sourceType="module"` and then set `classes` to false explicitly to achieve what you want.

caridy:
:+1:

nzakas:
I'll take a stab at the feature flag, then we can attack the other pieces.

xjamundx:
Oh I have PRs for the last 2 I can open I think...not sure how I missed this thread!

nzakas:
Working on tests for sourceType="module"

nzakas:
Wow, this is a really good test. Uncovered a few issues.

nzakas:
All done!

PR #73, opened 2015-03-05
New: Add modules feature flag (refs #72)

nzakas:


nzakas:
This basically switches to using `ecmaFeatures.modules` for enabling module features.

Still a bunch of tests missing (for classes, strict mode, etc.), but this should be a good foundation to build upon.

tests/lib/modules.js
```diff
@@ -1,148 +0,0 @@
-/**
```
xjamundx: Love putting the modules back in ecma-features!

nzakas: Really that was @caridy's original idea. Just took me a while to see that he was right. :)

xjamundx:
Do you want to sneak something like https://github.com/caridy/espree/pull/1in here as well?

nzakas:
Was hoping I could get you to do that after I merge this :)

xjamundx:
Sure!

nzakas:
Okay, all ready for you!

PR #74, opened 2015-03-06
Test: strict mode enforced in modules (refs #72)

xjamundx:
Not the most advanced tests, but should effectively ensure we're covered at the basic level. 

I believe the `import` statement is required because the way the `module` tests currently work in that they need to break when the module feature is off.


nzakas:
See #73 first, I rearranged things a bit

xjamundx:
Oh nice! Will close for now and wait to see what lands.

PR #75, opened 2015-03-06
Test: missing tests for modules & classes (refs #72)

xjamundx:
Added tests for the following:
- [x] `export default class extends bar {};`
- [x] `export default class {};`
- [x] `export default class foo extends bar {};`
- [x] `export default class foo {};`
- [x] `export class foo extends bar {};`
- [x] `export class foo {};`

Fixed the following:
- [x] Provided a fix for `export default class foo extends bar {};` such that it would not `throw`

Need to check the following:
- [x] AST for `export default class {};` in `class-default-anonymous.result.js`
- [x] AST for `export default class foo extends bar {};` in `class-default-anonymous-extends.result.js`

The ASTs are tricky and need to be verified as accurate, because they were generated by espree.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -4842,8 +4842,7 @@ function parseExportDefaultDeclaration() {
           declaration = parseFunctionExpression();
           return markerApply(marker, astNodeFactory.createExportDefaultDeclaration(declaration));
         } else if (lookahead.value === "class") {
-            // TODO: change this to declaration once we get ClassDeclaration with nullable name
-            parseClassExpression();
+            declaration = parseClassDeclaration();
             return markerApply(marker, astNodeFactory.createExportDefaultDeclaration(declaration));
         }
     }
@@ -5106,7 +5105,7 @@ function parseClassExpression() {
 }
 
 function parseClassDeclaration() {
-    var id, superClass = null, marker = markerCreate(),
+    var id = null, superClass = null, marker = markerCreate(),
         previousStrict = strict, classBody;
 
     // classes run in strict mode
@@ -5114,7 +5113,9 @@ function parseClassDeclaration() {
 
     expectKeyword("class");
 
-    id = parseVariableIdentifier();
+    if (lookahead.type === Token.Identifier) {
+        id = parseVariableIdentifier();
+    }
 
     if (matchKeyword("extends")) {
         lex();
--- a/tests/fixtures/ecma-features-mix/modules-and-classes/class-default-anonymous.result.js
+++ b/tests/fixtures/ecma-features-mix/modules-and-classes/class-default-anonymous.result.js
@@ -1,0 +1,90 @@
+module.exports = {
+    "type": "Program",
+    "body": [
+        {
+            "type": "ExportDefaultDeclaration",
+			"declaration": {
+				"type": "ClassDeclaration",
+				"id": null,
+				"superClass": null,
+				"body": {
+					"type": "ClassBody",
+					"body": [],
+					"range": [
+						21,
+						23
+					],
+					"loc": {
+						"start": {
+							"line": 1,
+							"column": 21
+						},
+						"end": {
+							"line": 1,
+							"column": 23
+						}
+					}
+				},
+				"range": [
+					15,
+					23
+				],
+				"loc": {
+					"start": {
+						"line": 1,
+						"column": 15
+					},
+					"end": {
+						"line": 1,
+						"column": 23
+					}
+				}
+			},
+            "range": [
+                0,
+                23
+            ],
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 0
+                },
+                "end": {
+                    "line": 1,
+                    "column": 23
+                }
+            }
+        },
+        {
+            "type": "EmptyStatement",
+            "range": [
+                23,
+                24
+            ],
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 23
+                },
+                "end": {
+                    "line": 1,
+                    "column": 24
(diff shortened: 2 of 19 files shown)


xjamundx:
This will clash with 731d54f54153fd1c696e4d4385188665666bfe2f a bit

caridy:
yes, that has to be fixed in espree and esprima.

xjamundx:
Thanks so much for your work on this @caridy Really amazing work! Do you want me to look into this last class issue?

nzakas:
Can you both yeah take a look at #73 for my changes? I'd like to use that as a bad before making further changes.

xjamundx:
Okay will look at fixing that last test tomorrow!

xjamundx:
Looking quickly at `export default class extends bar {};` to see if there's an easy fix.

xjamundx:
Initially appears to be related to `parseClassDeclaration()` vs `parseClassExpression()`. Only in the case of modules we do this with `parseSourceElement`:

```js
case "class":
	if (allowClasses) {
		return parseClassDeclaration();
	}
	break;
````

Inside of `parseClassDeclaration` we assume a name with this bit of code (which i think chokes up on finding `extends` instead):

```js
expectKeyword("class");
id = parseVariableIdentifier();
if (matchKeyword("extends")) {
   lex();
   superClass = parseLeftHandSideExpressionAllowCall();
}
```

nzakas:
What's your definition of "not working"? I'm not clear on the bug you're trying to fix.

caridy:
Here is a related issue: 

https://github.com/eslint/espree/blob/master/tests/fixtures/ecma-features/modules/export-default-function.result.js#L7

when doing `export default function () {}` that's not an expression as we have it today (my bad), it should be a function declaration with no name. Probably we should fix the function declaration parser first, to support that case (which is spec'd as HoistDeclaration), and then we can do the same for classes, where we can have a class that has no name but it is still a class declaration exported as default.

xjamundx:
@nzakas the issue is that `export default class extends foo {} ` would `throw`
@caridy yeah you know way more about the internal of how classes should work. the latest commit includes a workaround by parsing it as an expression. we could also make it a declarations with a `null` name. I have a hunch it's not the only class AST that isn't 100% perfect :)

Let me know what to change!

xjamundx:
Updated the PR description with clearer info and asks.

nzakas:
Ah, I see, so this is the unnamed function declaration discussion. Got it. (Ref: https://github.com/estree/estree/pull/38)

So the change here is to allow a `null` ID for both function and class declarations.

caridy:
@nzakas correct.

tests/fixtures/ecma-features-mix/modules-and-classes/class-default-anonymous.result.js
```diff
@@ -0,0 +1,54 @@
+module.exports = {
+    "type": "Program",
+    "body": [
+        {
+            "type": "ExportDefaultDeclaration",
+            "declaration": null,
```
xjamundx: I feel like this is missing a `declaration` :|

espree.js
```diff
@@ -5117,6 +5122,10 @@ function parseClassDeclaration() {
     classBody = parseClassBody();
     strict = previousStrict;
 
+    if (expression) {
+        return markerApply(marker, astNodeFactory.createClassExpression(id, superClass, classBody));
```
xjamundx: Sounds like the path is to remove this and just allow it to send in a `null` id to `createClassDeclaration` instead....

tests/fixtures/ecma-features-mix/modules-and-classes/class-default-anonymous.result.js
caridy: ha, that's a bug.

xjamundx: awesome i'll try to take a stab at this tonight :)

espree.js
nzakas: Yup!

xjamundx:
Okay took care of the `ClassDeclaration` stuff. I'm less familiar with the `FunctionDeclaration` stuff! I think the ASTs make sense now :)

xjamundx:
Okay, rebased and I think I broke everything again :-p 1 sec...

xjamundx:
Ok, should be good to go. Let me know if you need anything else on this (like FunctionDecl stuff)...

nzakas:
Cool, I'm hanging out in #eslint on Freenode as I try to get all of this merged and into ESLint

espree.js
```diff
@@ -5114,7 +5113,9 @@ function parseClassDeclaration() {
 
     expectKeyword("class");
 
-    id = parseVariableIdentifier();
+    if (lookahead.type === Token.Identifier) {
```
nzakas: This isn't quite right because `parseClassDeclaration()` normally *must* have an identifier. So you really need to pass in a flag specifying whether the ID is optional or required.

xjamundx: Let me know if you have a specific `class` declaration that breaks, at the moment I think this code-path may only be called in `export` statements, so maybe it works by accident? maybe.

xjamundx: I'm going to be out for most of the day taking kids to the zoo,  but may be able to help later!

nzakas: That's okay, I'll just merge and fix it. Thanks!

nzakas: Just FYI, the problem case is:

```js
class {}
```

This should throw an error because class declarations require names, but this change parsed it okay.

PR #76, opened 2015-03-06
Test: strict mode enforced in modules (refs #72)

xjamundx:
a few simple tests to ensure strict mode happens within modules


nzakas:
Nice!

PR #77, opened 2015-03-06
Test: disallow import/export in function body

xjamundx:
- 1 test for `import`
- 1 test for `export`

let me know if you think we ought to go into each of various permutations.
fat arrows might be a good idea as well.


nzakas:
Nah, this is fine. Thanks!

PR #83, opened 2015-03-07
Remove XJS note from Esprima-FB incompatibilities

lencioni:
The Esprima-FB project recently renamed all references of `XJS` to
`JSX`, so this should no longer be listed as an incompatibility.

- https://github.com/facebook/esprima/issues/83
- https://github.com/facebook/esprima/pull/85


nzakas:
Thanks!

PR #99, opened 2015-03-19
Fix: bad error in parsing invalid class setter (fixes #98)

### LINKED ISSUE #98: Error for "missing argument in class setters" fails to format
Using espree master, this code :
```js
espree.parse('class A { set foo(){} }', { ecmaFeatures: { classes: true }});
```
Throws :
```
Error: ASSERT: Message reference must be in range
    at assert (/home/ubuntu/workspace/espree.js:78:15)
    at /home/ubuntu/workspace/espree.js:2083:17
    at String.replace (native)
    at throwError (/home/ubuntu/workspace/espree.js:2080:29)
    at throwErrorTolerant (/home/ubuntu/workspace/espree.js:2106:20)
    at tryParseMethodDefinition (/home/ubuntu/workspace/espree.js:2438:17)
    at parseClassBody (/home/ubuntu/workspace/espree.js:4984:22)
    at parseClassDeclaration (/home/ubuntu/workspace/espree.js:5065:17)
    at parseSourceElement (/home/ubuntu/workspace/espree.js:5095:28)
    at parseSourceElements (/home/ubuntu/workspace/espree.js:5143:25)
```
I think `tryParseMethodDefinition` L2438 should be `throwErrorTolerant(lookahead, Messages.UnexpectedToken, lookahead.value);`

Marsup:
I had some troubles getting a test to pass without resorting to the double definition of `message` and `description`, if you could provide some guidance I'd fix it.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -2435,7 +2435,7 @@ function tryParseMethodDefinition(token, key, computed, marker) {
                 rest: null
             };
             if (match(")")) {
-                throwErrorTolerant(lookahead, Messages.UnexpectedToken);
+                throwErrorTolerant(lookahead, Messages.UnexpectedToken, lookahead.value);
             } else {
                 parseParam(options);
                 if (options.defaultCount === 0) {
--- a/tests/fixtures/ecma-features/classes/invalid-class-setter-declaration.result.js
+++ b/tests/fixtures/ecma-features/classes/invalid-class-setter-declaration.result.js
@@ -1,0 +1,7 @@
+module.exports = {
+    "index": 18,
+    "lineNumber": 1,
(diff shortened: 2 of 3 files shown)


nzakas:
This doesn't look quite right to me as we rarely use that third parameter.

Marsup:
The error is what I would expect and when looking for another usage of this function I found the same one so it felt right, what's the alternative?

nzakas:
I don't know, I just woke and haven't had time to look at this bug yet.

nzakas:
Okay, now that my head is a bit clearer, this looks good. Thanks!

PR #112, opened 2015-03-28
Update: Remove guardedHandlers/handlers from try (fixes #71)

### LINKED ISSUE #71: Remove TryStatement.handlers
Ref estree/estree#1:

```js
interface TryStatement <: Statement {
    // ....
    handler: CatchClause | null;
    handlers: [ CatchClause ]; // Removed
    // ....
}
```

The `handlers` array on `TryStatement` nodes was replaced with a single `handler`, which may be either a `CatchClause` or `null`.

#70 added support for the singular `handler` property. The next step is to remove `handlers`. This will be a breaking change.

nzakas:


btmills:
Isn't this a breaking change?

nzakas:
Yup. Already updated the commit summary.

btmills:
Ah, I missed that. :+1: LGTM.

nzakas:
Thanks. Going to let this sit until I'm sure we don't need another patch release.

PR #141, opened 2015-05-14
Fix: Override ecmaFeatures with options, don't replace them

Skalman:
**Problem before this fix:**
E.g. `ecmaFeatures.blockBindings` is `true` by default, but using options `{ecmaFeatures:{}}` would cause all features to be `undefined`, causing the example below to throw an error:

```js
var espree = require('espree');
espree.parse('let a = 1', {ecmaFeatures: {}}); // Throws 'Unexpected token let'
```

----

I didn't see how this affects `espree.tokenize()`, but I figured I might as well provide the fix there too.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -5272,7 +5272,9 @@ function tokenize(code, options) {
 
     // apply parsing flags
     if (options.ecmaFeatures && typeof options.ecmaFeatures === "object") {
-        extra.ecmaFeatures = options.ecmaFeatures;
+        Object.keys(options.ecmaFeatures).forEach(function(key) {
+            extra.ecmaFeatures[key] = options.ecmaFeatures[key];
+        });
     }
 
     try {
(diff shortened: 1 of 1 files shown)


nzakas:
This is working as intended for backwards compatibility with Esprima 1.2.2.

Skalman:
Thanks for the quick response. Esprima 1.2.2 didn't have the `ecmaFeatures` option, so I take it you mean that `ecmaFeatures:{}` is supposed to work like Esprima 1.2.2, disabling features that are on by default. To me it's counter-intuitive that `ecmaFeatures:null` and `ecmaFeatures:{}` do different things, and it's also counter-intuitive that `ecmaFeatures:{arrowFunctions:true}` will turn off `blockBindings`.

If this is indeed intended behavior, would it be okay to add a clarification to `README.md`:

> if specified, you must explicitly set all features you want to use, including features on by default

nzakas:
I don't mind adding docs, but I don't think that sentence is particularly clear.

Skalman:
Agreed - I find it difficult to explain current behavior well since it's unnatural to me. Is the following any clearer?

```js
// specify parsing features (default only has blockBindings: true)
// setting this option replaces the default values
ecmaFeatures: {
```

nzakas:
Lgtm

PR #147, opened 2015-06-01
Fix: Incomplete Switch Statement Hangs (Fixes #146)

### LINKED ISSUE #146: Incomplete switch statement causes espree to hang/crash
The following code will crash espree (so long as it's followed by a newline):
```js
switch(x) {  case 'case':  break;

```

It's very easy to reproduce:
```
> require('espree').parse("switch(x) {  case 'case':  break;\n")
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
[1]    63533 abort      node

xjamundx:
This fix follows a pattern used elsewhere in the codebase

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -4068,6 +4068,9 @@ function parseSwitchCase() {
             break;
         }
         statement = parseSourceElement();
+        if (typeof statement === "undefined") {
+            break;
+        }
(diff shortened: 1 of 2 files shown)


nzakas:
Lgtm

Issue #137, opened 2015-05-02
Comment attachment is different with ASI


lo1tuma:
Given this code example with semicolon:

```js
a(); // comment
```

then the following AST is generated.
```js
{
    "type": "Program",
    "body": [
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "CallExpression",
                "callee": {
                    "type": "Identifier",
                    "name": "a"
                },
                "arguments": []
            },
            "trailingComments": [
                {
                    "type": "Line",
                    "value": " comment"
                }
            ]
        }
    ],
    "comments": [
        {
            "type": "Line",
            "value": " comment"
        }
    ]
}
```

When the semicolon is omitted, then the `trailingComment` is attached to different node.

```js
a() // comment
```

```js
{
    "type": "Program",
    "body": [
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "CallExpression",
                "callee": {
                    "type": "Identifier",
                    "name": "a"
                },
                "arguments": [],
                "trailingComments": [
                    {
                        "type": "Line",
                        "value": " comment"
                    }
                ]
            }
        }
    ],
    "comments": [
        {
            "type": "Line",
            "value": " comment"
        }
    ]
}
```

As you can see, in the first AST the `trailingComment` is attached to the `ExpressionStatement` node, in the second AST the `trailingComment` is attached to the `CallExpression` node.

gyandeeps:
what behavior is expected here? case 1 or case 2.
fyi, esprima does it the same way.

nzakas:
That doesn't mean it's correct ;)

The intent is to see how possible or is to always return the first ast even without semicolons.

nzakas:
Fixed with 3.0.0.

PR #150, opened 2015-06-18
Update: Parsing and tokenizer enhancements for Acorn bridge

RReverser:
Enhanced support for parsing and tokenizing in Acorn bridge.

Got currently working (although with some known validation issues found in Acorn):

 - arrowFunctions ([currently allows duplicate param names in non-strict mode](https://github.com/marijnh/acorn/issues/274))
 - binaryLiterals
 - blockBindings
 - classes ~~([currently doesn't validate that setter should have exactly one argument](https://github.com/marijnh/acorn/issues/275))~~
 - defaultParams (caveat: gets also enabled via `ecmaFeatures.destructuring`)
 - forOf
 - generators
 - globalReturn
 - jsx
 - objectLiteralComputedProperties
 - objectLiteralDuplicateProperties ~~([__ proto __ duplication is currently allowed](https://github.com/marijnh/acorn/issues/276))~~
 - objectLiteralShorthandMethods ~~([currently supports deprecated SpiderMonkey feature like { method() 42 }](https://github.com/marijnh/acorn/issues/277))~~
 - objectLiteralShorthandProperties
 - octalLiterals
 - regexUFlag ~~([doesn't validate upper bound of code point](https://github.com/marijnh/acorn/issues/278))~~
 - regexYFlag
 - restParams (caveat: gets also enabled via `ecmaFeatures.destructuring`)
 - spread
 - superInFunctions (caveat: gets also enabled via `ecmaFeatures.classes`~~; also [currently allows super outside of functions and classes](https://github.com/marijnh/acorn/issues/279)~~)
 - templateStrings ~~([currently allows octal character escapes](https://github.com/marijnh/acorn/issues/280))~~
 - unicodeCodePointEscapes

Disabled (not working yet):
 - destructuring
 - modules

I've also added error bridge so that error objects are thrown in Espree-like format and have more info than Acorn gives natively. Although messages and positions where error is thrown obviously differ, so I've looked through those one by one, and in most cases Acorn seems to give more reasonable error position, so I decided to update tests.

Also, Esprima's positions for `TemplateElement`s included characters that belong to parent but not to `TemplateElement` itself, so I took decision to fix those tests instead of emulating [incorrect] Esprima's behavior.

Also added/fixed/enhanced(?) tokenizer bridge so that all the tests but one are passing; that one is related to incompatibility in template splitting logic between Acorn and Esprima; not sure how to fix it, so delayed for later.

Let me know if you have any questions. I know it's still incomplete and buggy, but all of the 10 issues in enabled parsing tests are known Acorn issues and I'll try to fix them in the nearest time (or, if someone's willing to help, you're very welcome).

P.S. I didn't squash commits yet for easier navigating through history (often required since it's WIP).


espree.js
```diff
@@ -64,18 +63,94 @@ function resetExtra() {
...
+    }
+
+    // TODO: remove when marijnh/acorn#281 is resolved
+    if (result.type === "AssignmentPattern") {
+        delete result.operator;
+    }
+
+    // hide acorn-specific properties from comparison
+    // but leave for internal needs:
+    Object.defineProperties(result, {
```
nzakas: I don't think we can keep these properties because they will end up exposed in ESLint rules and then we will be beholden to theses forever. Any way to remove them ?

espree.js
```diff
@@ -88,22 +163,104 @@ function esprimaFinishNode(result) {
...
+        case tt.jsxTagEnd:
+            return ecma.jsx;
+
+        default:
+            return true;
     }
+}
 
-    return eat.apply(this, arguments);
+pp.extend("finishNode", function(finishNode) {
```
nzakas: Ooh wrote, this is much nicer than what I was doing. :)

espree.js
```diff
@@ -368,8 +534,14 @@ function parse(code, options) {
         if (extra.loc) {
             acornOptions.locations = true;
         }
-    }
 
+        if (extra.ecmaFeatures.jsx) {
+            if (extra.ecmaFeatures.spread !== false) {
```
nzakas: I don't think this is necessary. In original Espree, we just check to see if jsx is true and we are inside a jsx element when an ellipsis token is found.

tests/lib/tokenize.js
```diff
@@ -132,16 +143,17 @@ describe("tokenize()", function() {
                 loc: true,
                 range: true
             });
-
+            ast = getRaw(ast);
             assert.deepEqual(ast.tokens, require("../fixtures/tokenize/template-string-embedded2-result.tokens.js"));
         });
 
-        it("should produce tokens when tokenizing template string with embedded expressions", function() {
+        it.only("should produce tokens when tokenizing template string with embedded expressions", function() {
```
nzakas: Oops?

nzakas:
This looks like an amazing start! Some questions that came to mind:

* Can you explain more about the TemplateElement issue? Didn't quite follow.
* it looks like a bunch of the tests have ranges that are off by one - any idea why that is?

I felt like I learned a ton from looking at this code, and definitely feel like this is the way we should be heading, so thanks for the effort!

espree.js
RReverser: Not really because Acorn and Acorn-JSX heavily rely on them for internal needs since instead of look-aheads they often use look-behind technique which means looking back at some already parsed nodes and talking some info, incl. offsets, from them.

espree.js
RReverser: I did that for now just to make spread tests work; this should be added to other "TODO"s for separate analysis of options.

tests/lib/tokenize.js
RReverser: Ouch, fixed :)

espree.js
RReverser: Thoughts about this a bit more - in fact, I do have some ideas about how to efficiently remove/hide them from final AST completely, but not sure if that will be worth efforts.

RReverser:
**UPD:** Resolved most of the found issues in Acorn (all except but one which requires much more thoughtful approach, see comment in issue).

RReverser:
@nzakas You're welcome! I'll answer your questions a bit later, need to leave now.

RReverser:
@nzakas So... regarding question **(1)** about TemplateElements. Let's take example from tests:

```javascript
`Hello ${b}. a + 5 = ${a + 5}`
```

Esprima tokenizes this as following:

```javascript
[
    { type: "Template",   value: "`Hello ${" },
    { type: "Identifier", value: "b" },
    { type: "Template",   value: "}. a + 5 = ${" },
    { type: "Identifier", value: "a" },
    { type: "Punctuator", value: "+" },
    { type: "Numeric",    value: "5" },
    { type: "Template",   value: "}`" }
]
```

And Acorn *(with currently implemented conversion)* returns instead:

```javascript
[
    { type: "Template",   value: "`" },          // tt.backQuote
    { type: "Template",   value: "Hello " },     // tt.template
    { type: "Template",   value: "${" },         // tt.dollarBraceL
    { type: "Identifier", value: "b" },          // tt.name
    { type: "Punctuator", value: "}" },          // tt.braceR
    { type: "Template",   value: ". a + 5 = " }, // tt.template
    { type: "Template",   value: "${" },         // tt.dollarBraceL
    { type: "Identifier", value: "a" },          // tt.name
    { type: "Punctuator", value: "+" },          // tt.plusMin
    { type: "Numeric",    value: "5" },          // tt.num
    { type: "Punctuator", value: "}" },          // tt.braceR
    { type: "Template",   value: "" },           // tt.template
    { type: "Template",   value: "`" }           // tt.backQuote
]
```

Personally to me, Acorn's format looks more reasonable, as it explicitly splits user-defined parts (template strings) from delimiters (`` ` ``, `${`, `}`) that are in fact separate tokens with own meanings and belong to parent (template) itself but not to strings nearby.

This gives more flexibility in validating user-defined parts and template punctuators separately as well as easier conversion to Estree AST.

---

Now, question **(2)**. Answer to it consists of two parts.

---

First one is highly coupled with explanation above: since Esprima joins punctuators and user-defined strings into single token, it also doesn't split them later in node locations, and thus TemplateElement nodes locations cover not just their values (template elements theirselves) but also punctuators that happen to be nearby.

That is, in template string like:

```javascript
`1${x}2${y}3`
```

* `TemplateElement("1")` covers location not just for `1` but for `` `1${ ``;
* `TemplateElement("2")` covers location not just for `2` but for `` }2${ ``;
* `TemplateElement("3")` covers location not just for `3` but for `` }3` ``.

Thus we are getting pretty inconsistent and unreliable locations that don't match contents of our `TemplateElement`s and don't reflect ranges of raw representations.

IMO those template punctuators are just like commas inside of `ArrayExpression` / `SequenceExpression` / `ObjectLiteral` or operators inside of `BinaryExpression` / `LogicalExpression` and so on, and should be treated as such - that is, their locations should belong to parent nodes and not to nearby elements.

---

Second part of question **(2)** is related to not-so-critical and, well, pretty rare incompatibility: when statement is the last one in the parent `BlockStatement` and it doesn't end with `;` (thus depends on ASI), Esprima adds whitespace after it and to the `}` as part of the statement while Acorn covers only statement iself and behaves like this whitespace belongs to the parent.

Example:
```javascript
{ x = 1 }
```

Acorn returns range for `ExpressionStatement` as: `[2, 7]`
Esprima returns: `[2, 8]`

Personally, I don't have any strong opinion on which one is correct in this case, since both might make sense. I guess this should be left to be clarified as part of CST initiative.

---

I hope this covers your questions. Let me know if you have more or need detailed clarifications on any of these parts.

nzakas:
Ah, thanks for the explanations, that's useful.

One of Espree's goals is to be a drop-in replacement for Esprima, and as such, I think it's important to keep things consistent even if there's disagreement with what is the correct way. We just have no way of knowing which of these details are being relied on and which are not. So, we need to see how hard it is to fix as many of the differences as possible. Even off-by-one locations could affect things, it's impossible to know. ( I already had to adjust the location of the Program node due to inconsistencies. )

btmills:
Regarding **(2b)**: I looked into the way Esprima handles whitespace with ASI as I was looking at #137, and the [ASI section](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-examples-of-automatic-semicolon-insertion) of the spec seemed to provide the best explanation, as the examples there show a semicolon being inserted immediately before the next token. So while Acorn's behavior would be great for #137, I could see an argument that Esprima's behavior is more in line with the spec.

RReverser:
Well, at least template thing looks not just like disagreement but rather a bug to me since it makes impossible to perform separate analysis for user-defined stuff (while I do agree about **(2b)**).

hzoo:
Yeah to get template strings working correctly in babel-eslint I had to transform the tokens and somehow it worked (as part of `acorn-to-esprima.js`)

https://github.com/babel/babel-eslint/issues/31#issuecomment-100912106

https://github.com/babel/babel-eslint/blob/bcd8e2322a1e59176218dc2398980631c2fd6c91/acorn-to-esprima.js#L76

nzakas:
@RReverser I disagree that this is a bug. String tokens contain the quotes as part of the token value, I don't see why the template literal delimiters are any different. 

In any event, this doesn't change the need to match what Esprima is doing. As with JavaScript engines, compatibility trumps ideal correctness. Maybe we can incorporate some of @hzoo's work to make it happen.

hzoo:
It looks like the tokens in acorn don't contain quotes unlike in esprima where it does.

```js
// Example
{ type: 'String',
  value: 'a',
  start: 8,
  end: 11,
  loc: { start: { line: 1, column: 8 }, end: { line: 1, column: 11 } },
  range: [ 8, 11 ] }
```
Made an issue with babel-jscs https://github.com/jscs-dev/babel-jscs/issues/1 because the rule for quotes is trying to get quote information from the string tokens. babel-eslint didn't have an issue since the ESLint rule was iterating over the `Literal` nodes instead which has node.raw

I guess the changes I made https://github.com/babel/babel-eslint/blob/bcd8e2322a1e59176218dc2398980631c2fd6c91/acorn-to-esprima.js#L244-L259 correspond to the off by one issues in loc.

Maybe can look into https://github.com/babel/babel-eslint/issues/24 again? I think some stuff in there needs to be removed though (flow, es7)?

nzakas:
@hzoo I was thinking about exposing something on Espree that could take an Acorn instance and run it as Espree, so babel could be plugged in. I thought then babel-eslint could use Espree as the first pass and eliminate the duplicate code in the projects.

RReverser:
> I disagree that this is a bug. String tokens contain the quotes as part of the token value, I don't see why the template literal delimiters are any different.

Because templates are different from strings - unlike regular literals, they have own internal structure and thus are more comparable with `ObjectExpression` or similar nodes since they only look similar to strings for the end-user, but are handled completely differently.

Did example with `` `1${x}2${y}3` `` above look right to you in the sense that offset of literal part implicitly impacts it's width too?

RReverser:
Also regarding quotes - I do agree that they belong to TemplateLiteral, but not to TemplateElement which is only static part of it, and thus has nothing to do with quotes.

nzakas:
Yeah, I think the template range stuff makes sense.

For the other range issues, here's what I'd propose: for any range differences not related to templates (I'm assuming there are some in there), can you revert the tests back to how they were? 

I think at that point we can merge in your changes and start looking at smoothing out more of the incompatibilities like tokens and locations and such. We can continue to iterate in the Acorn branch until everything is resolved.

How does that sound?

RReverser:
Sure, will do. There is really a few of those not related to templates but to expression statements, but I'm currently preparing for a talk, and will be able to get back to this only in a few days. If by any chance you or someone else could get to this earlier, feel free to send PR and I'll be happy to merge.

nzakas:
Cool - no rush. I'm still laid up and can only spend a limited amount of time at a computer, but if I get a spurt of energy, I may give it a shot.

nzakas:
I've updated the tests and am manually merging into the acorn branch.

RReverser:
Thanks! Hopefully this helped, although currently not having time to continue (again).

nzakas:
Yes, it helped a lot. I'll continue as I get stronger.

PR #152, opened 2015-06-20
Fix: "await" is a future reserved word (fixes #151)

### LINKED ISSUE #151: "await" is a reserved word in module context
[11.6.2.2 Future Reserved Words](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-future-reserved-words)

jrvidal:


tests/fixtures/ecma-features/modules/invalid-await.result.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    "index": 11,
+    "lineNumber": 1,
+    "column": 12,
+    "description": "Use of future reserved word in strict mode"
+};
```
ilyavolodin: Missing linebreak at the end.

tests/fixtures/ecma-features/modules/invalid-await.src.js
```diff
@@ -0,0 +1 @@
+export var await;
```
ilyavolodin: Same

lib/syntax.js
```diff
@@ -113,7 +113,7 @@ module.exports = {
         }
     },
 
-    isStrictModeReservedWord: function(id) {
+    isStrictModeReservedWord: function(id, ecmaFeatures) {
```
nzakas: Why pass in the whole object when only one flag is necessary?

jrvidal: You're right of course. I did it mostly for consistency with `isKeyword` and to avoid the boolean parameter code smell.

nzakas: Gotcha, OK by me.

nzakas:
@jrvidal can you add a line break at the end of those files?

jrvidal:
I think it's already done!

nzakas:
Got it. LGTM.

Issue #158, opened 2015-07-05
``leadingComments`` of anonymous class export are broken.


h13i32maru:
Hi! I use Espree in [ESDoc](https://github.com/h13i32maru/esdoc).

A ``leadingComments`` position are unexpected, when I parse following code with Espree.

```javascript
/**
 * this is anonymous class.
 */
export default class {
  /**
   * this is method1.
   */
  method1(){
  }
}
```
<img width="486" alt="2015-07-05 12 06 17" src="https://cloud.githubusercontent.com/assets/985170/8510151/49c8d898-230e-11e5-8bff-99a130e48cb3.png">

(full JSON https://gist.github.com/h13i32maru/f22536ca8c224e08cc9e)

Espree option:
```javascript
  {
      comments: true,
      attachComment: true,
      loc: true,
      ecmaFeatures: {
        arrowFunctions: true,
        blockBindings: true,
        destructuring: true,
        regexYFlag: true,
        regexUFlag: true,
        templateStrings: true,
        binaryLiterals: true,
        octalLiterals: true,
        unicodeCodePointEscapes: true,
        defaultParams: true,
        restParams: true,
        forOf: true,
        objectLiteralComputedProperties: true,
        objectLiteralShorthandMethods: true,
        objectLiteralShorthandProperties: true,
        objectLiteralDuplicateProperties: true,
        generators: true,
        spread: true,
        classes: true,
        modules: true,
        jsx: true,
        globalReturn: true
      }
  };
```

Thank you.

nzakas:
Can you explain what's wrong? I can't tell from your screenshot.

h13i32maru:
Sorry, I explain it.

``this is anonymous class.`` is commented on ``export default class``, but the comment is inserted into ``leadingComments`` of ``method1()``.

<img width="486" alt="49c8d898-230e-11e5-8bff-99a130e48cb3" src="https://cloud.githubusercontent.com/assets/985170/8537537/56e2099e-249b-11e5-9060-21f3dc6bddfb.png">



Is this would be the correct behavior?
I expect that the comment is inserted into ``leadingComments`` of ``ExportDefaultDeclaration``.

nzakas:
Thanks for explaining. Yeah, that looks like a bug to me.

mysticatea:
I will try to fix this.

nzakas:
Thanks @mysticatea

h13i32maru:
@mysticatea Thank you :sparkles:

Issue #155, opened 2015-06-24
Comment Attachment is confused by comment within condintion


lo1tuma:
Given the following code:

```js
/* foo */
if (/* bar */  a) {}
```

produces the following partial AST when `attachComment` is enabled:

```json
{
  "type": "IfStatement",
  "test": {
    "type": "Identifier",
    "name": "a",
    "leadingComments": [
      {
        "type": "Block",
        "value": " foo "
      },
      {
        "type": "Block",
        "value": " bar "
      }
    ]
  },
  "consequent": {
    "type": "BlockStatement",
    "body": []
  },
  "alternate": null
}
```

As you can see both comments are attached to the `Identifier` node of the `IfStatement` test.

If the comment in the test condition is not present, the first comment `foo` is directly attached to the `IfStatement` node. 

```js
/* foo */
if (a) {}
```

```json
{
  "type": "IfStatement",
  "test": {
    "type": "Identifier",
    "name": "a"
  },
  "consequent": {
    "type": "BlockStatement",
    "body": []
  },
  "alternate": null,
  "leadingComments": [
    {
      "type": "Block",
      "value": " foo "
    }
  ]
}
```

I would expect a more reliable behavior of the comment attachment algorithm, so that the first comment `foo` is always attached to the same node (preferable the `IfStatement` node).

Same problem exists for `WhileStatement` and I would guess `ForStatement`, `ForInStatement`, and `ForOfStatement`.

nzakas:
Agreed, seems like a bug. Comment attachment is hard.

PR #162, opened 2015-07-10
Fix: Leading comments for anonymous classes (fixes: #158)

mysticatea:
The comment attachment process looks below:

1. Leaf-nodes (A) gets comments that are before itself.
2. Containing-nodes gets comments from A if the last leading comment of A is before itself.

In this (#158) case, since the `ClassDeclaration` does not have its `Identifier`, the `Identifier` of the first `MethodDefinition` gets the comment of the class.
And the `Identifier` gets also the comment of the method.
Next, the `MethodDefinition` gets the comments from the `Identifier`.

The `ClassBody` node does not get the comments from the `MethodDefinition`, because the last leading comment of the `MethodDefinition` is not before the `ClassBody` node. But the comment of the class is in the leading comments of the `MethodDefinition`.

So, I modified step 2 of the process.

* It finds comments that are before containing-nodes from the leading comment list of A, and the containing-node gets those.


mysticatea:
This PR probably fixes #155, but I didn't write a test for it.

Please wait...

mysticatea:
OK, I added a test for #155.
It's fixed :)

nzakas:
LGTM

PR #166, opened 2015-07-15
New: Add experimental object rest/spread (fixes #163)

### LINKED ISSUE #163: Support experimental object spread
Object spread is an experimental feature, however, React encourages its use and as such it is one of the most commonly requested features. While this does not mean we will start supporting experimental features, we will make an exception because this keeps coming up as holding back React users from using ESLint out of the box.

The feature should be named `experimentalObjectRestSpread` and allow this syntax:

```
let { ...foo } = bar;
let foo = { a, ...b };
```

Full spec: https://github.com/sebmarkbage/ecmascript-rest-spread

nzakas:


README.md
```diff
@@ -126,7 +129,10 @@ var ast = espree.parse(code, {
         jsx: true,
 
         // enable return in global scope
-        globalReturn: true
+        globalReturn: true,
+
+        // allow experimental object rest/spread
+        experimentalObjectRestSpread: true
```
gyandeeps: Would this be better as an object like this:

```js
experimental: {
       ObjectRestSpread: true
}
```

So that its easy to add more and they are all under the same blovk

nzakas: No, that would complicate the logic. Better to just keep it a single name like everything else.

PR #186, opened 2015-08-11
Fix: trailing commas in import/export (fixes #148)

### LINKED ISSUE #148: Support for trailing commas in Import/Export statements
Babel supports this, I need to check with the spec:

- `import {y,} from 'x';` 
- `export {y,} from 'x';`

xjamundx:


espree.js
```diff
@@ -4771,7 +4771,7 @@ function parseExportNamedDeclaration() {
         do {
             isExportFromIdentifier = isExportFromIdentifier || matchKeyword("default");
             specifiers.push(parseExportSpecifier());
-        } while (match(",") && lex());
+        } while (match(",") && lex() && !match("}"));
```
nzakas: This seems a little fragile to me. What if there's a `+` after an identifier?

xjamundx: Okay let me try to find some cases that cause it to break and go from there. I don't think it will be too difficult to fix either way...

xjamundx: So far I could not break it. Let me know if you have any more suggestions!

nzakas: Ok then, let's go with this.

PR #223, opened 2015-12-09
Breaking: move ecmaFeatures into ecmaVersion (fixes #222)

### LINKED ISSUE #222: Add ecmaVersion, eliminate most ecmaFeatures
We introduced `ecmaFeatures` as a way to start incrementally implementing ECMAScript 6 at a time when it was constantly changing. We did this in order to allow people to start using those features that were ready without forcing usage of all features.

Now that ES6 is finalized and all the features are implemented, most people are using all ES6 features and it's a pain to have to manually configure them all. Additionally, there's no good reason to turn on just some features and not all of them (as that's not how the real world works).

To move forward, we'll be making the following changes:

1. Introduce `ecmaVersion` as the way to indicate which version of ECMAScript you want to target. Possible values will be 3, 5, and 6. This will pass through to Acorn.
1. Remove all ES6 flags from `ecmaFeatures` (`globalReturn`, `jsx`, and experimental features will remain)
1. Remove individual checks for ES6 features (could potentially improve performance).

In all likelihood, going forward we will probably follow a similar pattern of introducing experimental features using `ecmaFeatures` and then rolling them into a larger `ecmaVersion` grouping later on.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28988665-add-ecmaversion-eliminate-most-ecmafeatures?utm_campaign=plugin&utm_content=tracker%2F9348450&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F9348450&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nzakas:


tests/fixtures/ecma-features-mix/modules-and-blockBindings/export-let-object.config.js
```diff
@@ -1,4 +0,0 @@
-module.exports = {
```
xjamundx: yay being able to get rid of all this config stuff in the tests!

nzakas: :+1:

PR #228, opened 2016-01-03
New: ecmaFeatures.impliedStrict (fixes: #227)

nre:
- New feature `ecmaFeatures.impliedStrict`.
- README.
- Two tests.
- Modified tests/lib/ecma-features.js:
  - `ecmaFeatures[feature] = true` throws for `impliedStrict`.

### DIFF
--- a/tests/fixtures/ecma-features/impliedStrict/inner-strict-violation.src.js
+++ b/tests/fixtures/ecma-features/impliedStrict/inner-strict-violation.src.js
@@ -1,0 +1,3 @@
+function inner() {
+    var private;
+};
--- a/tests/lib/ecma-features.js
+++ b/tests/lib/ecma-features.js
@@ -57,6 +57,15 @@ var testFiles = shelljs.find(FIXTURES_DIR).filter(function(filename) {
 // Tests
 //------------------------------------------------------------------------------
 
+/**
+ * Returns whether a feature should throw in its tests when it is enabled.
+ * @param {string} feature The name of the feature.
+ * @returns {boolean} Whether it should throw in its tests when it is enabled.
+ */
+function shouldThrowInTestsWhenEnabled(feature) {
+    return (feature === "impliedStrict");
+}
+
 describe("ecmaFeatures", function() {
 
     var config;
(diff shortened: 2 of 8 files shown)


tests/lib/ecma-features.js
```diff
@@ -75,17 +75,18 @@ describe("ecmaFeatures", function() {
         // Uncomment and fill in filename to focus on a single file
         // var filename = "jsx/invalid-matching-placeholder-in-closing-tag";
         var feature = path.dirname(filename),
+            permissiveValue = feature !== "impliedStrict",
```
nzakas: Why this change?

tests/fixtures/ecma-features/impliedStrict/inner-strict-violation.src.js
```diff
@@ -0,0 +1,3 @@
+function inner() {
+    var private;
```
nzakas: This should throw a parsing error, correct? If so, the result file should have an error object, not an AST.

tests/lib/ecma-features.js
nzakas: Ooh, I think you're doing this because the implied strict tests are failing. Please revert this change and fix the tests.

nre:
The reason I changed ecma-features.js is that I don't know any errors that are silenced by strict mode. In other words, what code will make a 'non-strict' parser throw, but not a 'strict' parser? So for `impliedStrict` I swapped the 'correct/error' roles of the true/false feature settings.

The other ecmaFeatures,

- `jsx`
- `globalReturn`
- `experimentalObjectRestSpread`

all solve parser errors when they are `true`. But `impliedStrict` finds more errors when it is `true`.

The `impliedStrict` tests generate the ASTs in the result files when `impliedStrict: false`, and throw when `impliedStrict: true`. This is the opposite of the other ecmaFeatures tests, but I don't think that means they are failing.

nzakas:
Ah, I see, thanks for explaining. We should split out that case in a more obvious way, then. Maybe create an array called `failOnEnable` and put `impliedStrict` in it?

nre:
I moved the condition inside `shouldThrowInTestsWhenEnabled(feature)`.

nzakas:
Fair enough, thanks!

PR #239, opened 2016-01-19
Fix: tests for importing keywords (fixes #225)

### LINKED ISSUE #225: Should not allow reserved words for variables which are declared by import declarations
```js
import {class, default, var} from "foo";
```

Currently, espree succeeds to parse this code.
But should raise a syntax error.

http://tc39.github.io/ecma262/#prod-ImportSpecifier

```
ImportSpecifier:
    ImportedBinding
    IdentifierName "as" ImportedBinding

ImportedBinding:
    BindingIdentifier

BindingIdentifier[Yield]:
    Identifier
    [~Yield] "yield"

Identifier::
    IdentifierName but not ReservedWord
```

Reserved words are allowed only where followed by "as".

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29132323-should-not-allow-reserved-words-for-variables-which-are-declared-by-import-declarations?utm_campaign=plugin&utm_content=tracker%2F9348450&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F9348450&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mysticatea:
fixes #225.

I fixed this in https://github.com/ternjs/acorn/commit/327d631aa38c75c1d8a97bcb6ba3aca4a148464f
So I just add tests for this.

The test of `ecma-version/6/modules/invalid-import-named-keyword` had been failing with `acorn@2.6.4`.
Then the test gets passing with `acorn@2.7.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   },
   "license": "BSD-2-Clause",
   "devDependencies": {
+    "acorn": "^2.7.0",
     "browserify": "^7.0.0",
     "chai": "^1.10.0",
     "complexity-report": "~0.6.1",
--- a/tests/fixtures/ecma-version/6/modules/import-named-as-specifier-keyword.result.js
+++ b/tests/fixtures/ecma-version/6/modules/import-named-as-specifier-keyword.result.js
@@ -1,0 +1,275 @@
+module.exports = {
+    "type": "Program",
+    "loc": {
+        "start": {
+            "line": 1,
+            "column": 0
+        },
+        "end": {
+            "line": 1,
+            "column": 31
+        }
+    },
+    "range": [
+        0,
+        31
+    ],
+    "body": [
+        {
+            "type": "ImportDeclaration",
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 0
+                },
+                "end": {
+                    "line": 1,
+                    "column": 31
+                }
+            },
+            "range": [
+                0,
+                31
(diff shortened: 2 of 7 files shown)


mysticatea:
CI would recover after #236.

package.json
```diff
@@ -31,6 +31,7 @@
   },
   "license": "BSD-2-Clause",
   "devDependencies": {
+    "acorn": "^2.7.0",
```
gyandeeps: This is already present under dependencies.

tests/fixtures/ecma-version/6/modules/import-named-as-specifier-keyword.result.js
```diff
@@ -0,0 +1,275 @@
+module.exports = {
```
ilyavolodin: Might want to add Tern-JS copyright at the top, since this is copied from their repository

nzakas: No need, this isn't executable code.

package.json
nzakas: I'll fix this manually.

mysticatea: aaa, I'm sorry. Thank you!

Issue #212, opened 2015-11-04
Incorrect trailing comments


nzakas:
Given this code:

```js
//foo
var zzz /*aaa*/ = 777;
//bar
```

Espree incorrectly appends `//foo` as a trailing comment on the `Literal` 777.

kaicataldo:
I'm going to take a look at this now that the leading comments bug is fixed

kaicataldo:
This doesn't seem to be an issue any more. However, wanted to confirm that the current behavior of leading comments is correct. Namely, wondering if `//foo` should be appending to the leading comments of the Literal 777.

```js
{
    "type": "Program",
    "start": 0,
    "end": 34,
    "range": [
        6,
        28
    ],
    "body": [
        {
            "type": "VariableDeclaration",
            "start": 6,
            "end": 28,
            "range": [
                6,
                28
            ],
            "declarations": [
                {
                    "type": "VariableDeclarator",
                    "start": 10,
                    "end": 27,
                    "range": [
                        10,
                        27
                    ],
                    "id": {
                        "type": "Identifier",
                        "start": 10,
                        "end": 13,
                        "range": [
                            10,
                            13
                        ],
                        "name": "zzz",
                        "trailingComments": [
                            {
                                "type": "Block",
                                "value": "aaa",
                                "start": 14,
                                "end": 21,
                                "range": [
                                    14,
                                    21
                                ]
                            }
                        ]
                    },
                    "init": {
                        "type": "Literal",
                        "start": 24,
                        "end": 27,
                        "range": [
                            24,
                            27
                        ],
                        "value": 777,
                        "raw": "777",
                        "leadingComments": [
                            {
                                "type": "Line",
                                "value": "foo",
                                "start": 0,
                                "end": 5,
                                "range": [
                                    0,
                                    5
                                ]
                            },
                            {
                                "type": "Block",
                                "value": "aaa",
                                "start": 14,
                                "end": 21,
                                "range": [
                                    14,
                                    21
                                ]
                            }
                        ]
                    }
                }
            ],
            "kind": "var",
            "leadingComments": [
                {
                    "type": "Line",
                    "value": "foo",
                    "start": 0,
                    "end": 5,
                    "range": [
                        0,
                        5
                    ]
                }
            ],
            "trailingComments": [
                {
                    "type": "Line",
                    "value": "bar",
                    "start": 29,
                    "end": 34,
                    "range": [
                        29,
                        34
                    ]
                }
            ]
        }
    ],
[...]

PR #257, opened 2016-03-12
Fix: leading comments added from previous node (fixes #256)

### LINKED ISSUE #256: Incorrect leading comments.
From https://github.com/eslint/eslint/pull/5512

In the following case:

```js
var foo = {
  // foo
};
var bar;
```

`// foo` comment has attached to the leading of `var bar`.

----

```js
"use strict";

const espree = require("espree");
const ast = espree.parse(
    `var foo = {
  // foo
};
var bar;`,
    {
        range: true,
        attachComment: true
    }
);

console.log(JSON.stringify(ast, null, 4));
```

```json
{
    "type": "Program",
    "start": 0,
    "end": 32,
    "range": [
        0,
        32
    ],
    "body": [
        {
            "type": "VariableDeclaration",
            "start": 0,
            "end": 23,
            "range": [
                0,
                23
            ],
            "declarations": [
                {
                    "type": "VariableDeclarator",
                    "start": 4,
                    "end": 22,
                    "range": [
                        4,
                        22
                    ],
                    "id": {
                        "type": "Identifier",
                        "start": 4,
                        "end": 7,
                        "range": [
                            4,
                            7
                        ],
                        "name": "foo"
                    },
                    "init": {
                        "type": "ObjectExpression",
                        "start": 10,
                        "end": 22,
                        "range": [
                            10,
                            22
                        ],
                        "properties": [],
                        "trailingComments": [
                            {
                                "type": "Line",
                                "value": " foo",
                                "start": 14,
                                "end": 20,
                                "range": [
                                    14,
                                    20
                                ]
                            }
                        ]
                    }
                }
            ],
            "kind": "var"
        },
        {
            "type": "VariableDeclaration",
            "start": 24,
            "end": 32,
            "range": [
                24,
                32
            ],
            "declarations": [
                {
                    "type": "VariableDeclarator",
                    "start": 28,
                    "end": 31,
                    "range": [
                        28,
                        31
                    ],
                    "id": {
                        "type": "Identifier",
                        "start": 28,
                        "end": 31,
                        "range": [
                            28,
                            31
                        ],
                        "name": "bar"
                    },
                    "init": null
                }
            ],
            "kind": "var",
            "leadingComments": [
                {
                    "type": "Line",
                    "value": " foo",
                    "start": 14,
                    "end": 20,
                    "range": [
                        14,
                        20
                    ]
                }
            ]
        }
    ],
    "sourceType": "script",
    "comments": [
        {
            "type": "Line",
            "value": " foo",
            "start": 14,
            "end": 20,
            "range": [
                14,
                20
            ]
        }
    ]
}
```

kaicataldo:
Looks like the `mootools` test is failing. Not sure if this means my change is bad or the expected output from the test is incorrect now that this change has been made. Are these generated with a command somewhere?

I ran this version of my change against the current master branch to test for speed and was surprised to find no conclusive evidence of this slowing things down (most likely because my change only runs on comments and only when the node type is `Program`)...Some tests were a bit slower, and others were a bit faster. So doesn't seem like a huge impact on performance :+1:

### DIFF
--- a/tests/fixtures/attach-comments/comment-within-previous-node.result.js
+++ b/tests/fixtures/attach-comments/comment-within-previous-node.result.js
@@ -1,0 +1,463 @@
+module.exports = {
+    "type": "Program",
+    "loc": {
+        "start": {
+            "line": 1,
+            "column": 0
+        },
+        "end": {
+            "line": 6,
+            "column": 8
+        }
+    },
+    "range": [
+        0,
+        54
+    ],
+    "body": [
+        {
+            "type": "VariableDeclaration",
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 0
+                },
+                "end": {
+                    "line": 4,
+                    "column": 2
+                }
+            },
+            "range": [
+                0,
+                44
+            ],
+            "declarations": [
+                {
+                    "type": "VariableDeclarator",
+                    "loc": {
+                        "start": {
+                            "line": 1,
+                            "column": 4
+                        },
+                        "end": {
+                            "line": 4,
+                            "column": 1
+                        }
+                    },
+                    "range": [
+                        4,
+                        43
+                    ],
+                    "id": {
+                        "type": "Identifier",
+                        "loc": {
+                            "start": {
+                                "line": 1,
+                                "column": 4
+                            },
+                            "end": {
+                                "line": 1,
+                                "column": 7
+                            }
+                        },
+                        "range": [
+                            4,
+                            7
+                        ],
+                        "name": "foo"
+                    },
+                    "init": {
+                        "type": "ObjectExpression",
+                        "loc": {
+                            "start": {
+                                "line": 1,
+                                "column": 10
+                            },
+                            "end": {
+                                "line": 4,
+                                "column": 1
+                            }
+                        },
+                        "range": [
+                            10,
+                            43
+                        ],
+                        "properties": [
+                            {
+                                "type": "Property",
+                                "loc": {
+                                    "start": {
+                                        "line": 2,
+                                        "column": 4
+                                    },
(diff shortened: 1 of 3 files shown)


nzakas:
Can you add a test for this change?

Chances are that there was an error in Mootools output that this fixes. You can run `node tools/update-tests.js` to update, though you should revert any other changes that tool makes as it causes some errors in other tests (haven't gotten around to fixing that).

kaicataldo:
Yeah, of course! Sorry, wasn't understanding how these tests were organized/generated, but I see now. Let me know if I can do anything else!

May take a look at #212 after this PR lands, since I've familiarized myself with the code and the change is probably a similar thing :)

tests/fixtures/attach-comments/comment-within-previous-node.result.js
```diff
@@ -0,0 +1,463 @@
...
+                                    },
+                                    "range": [
+                                        21,
+                                        26
+                                    ],
+                                    "value": "bar",
+                                    "raw": "'bar'"
+                                },
+                                "kind": "init",
+                                "trailingComments": [
```
kaicataldo: Correctly adds trailing comment to `bar` property

tests/fixtures/attach-comments/comment-within-previous-node.result.js
```diff
@@ -0,0 +1,463 @@
...
+                    "column": 8
+                }
+            },
+            "range": [
+                46,
+                54
+            ],
+            "declarations": [
+                {
+                    "type": "VariableDeclarator",
```
kaicataldo: No leading comments!

nzakas:
Awesome, thank you!

PR #262, opened 2016-03-16
Fix: Fix behavior of ignoring comments within previous nodes (refs #256)

kaicataldo:
While adding tests to the main ESLint repo to account for the changes I made in https://github.com/eslint/espree/pull/257 I realized that some of my initial assumptions were wrong:
 - That the `body` prop is always an array (in reality it can be an array or an object)
 - I didn't check the `body` props recursively (I had incorrectly thought the `Program` node's `body` contained all the nodes, which, in retrospect, is pretty silly)

This means that my current merged fix does not account for this case, as the comment to remove is nested a few levels deep, with the `body` props being a mixture of arrays and objects as you drill down:
```js
function a() {
    var b = {
        // comment
    };
    return b;
}
```
I'm going to test out the main ESLint repo with this branch now, but wanted to see if I could get some eyes on this to see if it made sense. Thanks!

### DIFF
--- a/lib/comment-attachment.js
+++ b/lib/comment-attachment.js
@@ -43,6 +43,40 @@ var extra = {
     bottomRightStack: []
 };
 
+/** Recursively remove leading comments that are incorrectly added when no
+ * expression exists between comment and the current node
+ * @param {node} node to recursively check
+ * @returns {void}
+ */
+function removeExtraLeadingComments(node) {
+    var i, j;
+
+    if (!node.body) {
+        return;
+    }
+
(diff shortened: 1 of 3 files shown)


kaicataldo:
Tested by `npm link`ing it to the main eslint repo. Was able to remove [this guard](https://github.com/eslint/eslint/pull/5512/files#diff-34bea6f5f11bfaebdf332e35b4f14ae6R92) (which is what alerted me to this problem in the first place!) and all the tests pass.

nzakas:
Lgtm

PR #265, opened 2016-03-27
Fix: Disallow namespaces objects in JSX (fixes #261)

### LINKED ISSUE #261: Disallow namespace objects in JSX
We were parsing the following invalid JSX without error:

```jsx
<a.b:c></a.b:c>;
```

This was due to an issue in Acorn-JSX that has now been resolved: https://github.com/RReverser/acorn-jsx/issues/27

We should update to ensure that we no longer parse this syntax.

kaicataldo:
Fixed `invalid-namespace-value-with-dots.result.js`, which is a pre-existing test but for some reason had an AST it was testing against rather than an error message, as it should have (which is why we didn't catch this issue).

 Renamed the `invalid-namespace-name-with-docts` tests for better clarity and consistency with the other tests I added.


nzakas:
LGTM. Thanks!

PR #266, opened 2016-03-31
Build: replace optimist with a simple native method

shinnn:
Because:

1. [`optimist`](https://www.npmjs.com/package/optimist) package was [deprecated](https://github.com/substack/node-optimist/commit/815c965cee7682ae2c6a80f8b306d20c81eb199a)  long ago.
2. It can be replaced with a simple, enough readable native method.

Otherwise, I recommend to use [minimist](https://www.npmjs.com/package/minimist) instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "json-diff": "~0.3.1",
     "leche": "^1.0.1",
     "mocha": "^2.0.1",
-    "optimist": "~0.6.0",
     "regenerate": "~0.5.4",
     "shelljs": "^0.3.0",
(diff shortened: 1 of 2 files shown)


nzakas:
Can you please update your commit message to begin with "Build:"?

shinnn:
Updated.

shinnn:
Ah, sorry. Is the ESLint contributing guide applied to all the ESLint-related projects? If so, I should have opened an issue before submitting this patch.

nzakas:
Yup, same guidelines. It's OK this time. :)

PR #269, opened 2016-04-18
Fix: remove extra leading comments at node level (fixes #264)

### LINKED ISSUE #264: Leading comments incorrectly duplicated on variable initialization
Leading comments are incorrectly duplicated on variable initializations.  The current strategy employed to fix https://github.com/eslint/espree/issues/256 only checks the array inside `body` properties and removes comments that exist in the range of the previous index's node, never checking the `init` property (and so not covering this case). Just to be clear, this is not a regression - this is just one more case that was being added incorrectly (or rather, in practice, not being removed correctly). 

The difficulty with this one is that when we drill deeper into the tree, we'll need to have saved some kind of reference to the previously iterated over comments to know if we need to remove it or not. We may have to rethink our strategy on this.

Example:
The following code
```js
//foo
var zzz /*aaa*/ = 777;
//bar
```
currently produces this output (notice the incorrect leading comment `//foo` on the literal `777`):
```js
{
    "type": "Program",
    "start": 0,
    "end": 34,
    "range": [
        6,
        28
    ],
    "body": [
        {
            "type": "VariableDeclaration",
            "start": 6,
            "end": 28,
            "range": [
                6,
                28
            ],
            "declarations": [
                {
                    "type": "VariableDeclarator",
                    "start": 10,
                    "end": 27,
                    "range": [
                        10,
                        27
                    ],
                    "id": {
                        "type": "Identifier",
                        "start": 10,
                        "end": 13,
                        "range": [
                            10,
                            13
                        ],
                        "name": "zzz",
                        "trailingComments": [
                            {
                                "type": "Block",
                                "value": "aaa",
                                "start": 14,
                                "end": 21,
                                "range": [
                                    14,
                                    21
                                ]
                            }
                        ]
                    },
                    "init": {
                        "type": "Literal",
                        "start": 24,
                        "end": 27,
                        "range": [
                            24,
                            27
                        ],
                        "value": 777,
                        "raw": "777",
                        "leadingComments": [
                            {
                                "type": "Line",
                                "value": "foo",
                                "start": 0,
                                "end": 5,
                                "range": [
                                    0,
                                    5
                                ]
                            },
                            {
                                "type": "Block",
                                "value": "aaa",
                                "start": 14,
                                "end": 21,
                                "range": [
                                    14,
                                    21
                                ]
                            }
                        ]
                    }
                }
            ],
            "kind": "var",
            "leadingComments": [
                {
                    "type": "Line",
                    "value": "foo",
                    "start": 0,
                    "end": 5,
                    "range": [
                        0,
                        5
                    ]
                }
            ],
            "trailingComments": [
                {
                    "type": "Line",
                    "value": "bar",
                    "start": 29,
                    "end": 34,
                    "range": [
                        29,
                        34
                    ]
                }
            ]
        }
    ],
    "sourceType": "script",
    "comments": [
        {
            "type": "Line",
            "value": "foo",
            "start": 0,
            "end": 5,
            "range": [
                0,
                5
            ]
        },
        {
            "type": "Block",
            "value": "aaa",
            "start": 14,
            "end": 21,
            "range": [
                14,
                21
            ]
        },
        {
            "type": "Line",
            "value": "bar",
            "start": 29,
            "end": 34,
            "range": [
                29,
                34
            ]
        }
    ]
}
```

I plan to look at this.

kaicataldo:
I think I've updated all the library tests correctly (using tools/update-tests.js). Before we decide to merge, please see #264 for the failing test in the ESLint repo. If all looks good, I'll update that test and update the espree dependency in ESLint then.

---

This PR changes the recursive check and removal of extra leading comments at the `Program` node level in favor of saving off the node previously processed for comment attachment and checking the leading comments' respective locations against the previously saved node's to ensure they are correct before attachment.

Both implementations are doing essentially the same check, but the changes in this PR allows us to do the check at the time of comment attachment at the individual node level (and fixes some cases my original recursive check wasn't catching), rather than do an extra traversal of the entire tree at the end and try to remove unwanted leading comments after the fact (an error prone procedure - we've seen a few cases my original change didn't catch).


tests/fixtures/attach-comments/mix-line-and-block-comments.result.js
```diff
@@ -106,24 +106,6 @@ module.exports = {
                         "raw": "777",
                         "leadingComments": [
                             {
-                                "type": "Line",
```
kaicataldo: Incorrect leading comment

tests/fixtures/attach-comments/switch-no-default-comment-in-nested-functions.result.js
```diff
@@ -649,27 +649,7 @@ module.exports = {
                                                 ],
                                                 "value": false,
                                                 "raw": "false"
-                                            },
```
kaicataldo: Incorrectly attached leading comment

nzakas:
This looks great! Thanks!

PR #276, opened 2016-05-26
Convert ~ and ! prefix tokens to esprima

### LINKED ISSUE #274: Exclamation mark has acorn token type
When having this sample code:

```
if (!true) {}
```

The token for the exclamation mark is an acorn token type object and not a string `Punctator`. I wasn't sure how to fix, as acorn has no token type for `!`. Simple string compare of the value?

```
...
{
      "type": "Punctuator",
      "value": "(",
      "start": 3,
      "end": 4,
      "range": [
        3,
        4
      ]
    },
    {
      "type": {
        "label": "prefix",
        "beforeExpr": true,
        "startsExpr": true,
        "isLoop": false,
        "isAssign": false,
        "prefix": true,
        "postfix": false,
        "binop": null,
        "updateContext": null
      },
      "value": "!",
      "start": 4,
      "end": 5,
      "range": [
        4,
        5
      ]
    },
    {
      "type": "Boolean",
      "value": "true",
      "start": 5,
      "end": 9,
      "range": [
        5,
        9
      ]
    },
...
```

danez:
The two prefixes are now converted to Punctator like esprima is returning them.

 fixes #274


lib/token-translator.js
```diff
@@ -123,8 +123,9 @@ TokenTranslator.prototype = {
                  type === tt.arrow || type === tt.jsxTagStart ||
                  type === tt.incDec || type === tt.starstar ||
                  type === tt.jsxTagEnd || (type.binop && !type.keyword) ||
-                 type.isAssign) {
-
+                 type.isAssign ||
+                 (type.prefix && (token.value === "~" || token.value === "!"))
```
nzakas: Do we need to filter further if `type.prefix` is true? Is there ever a situation where a prefix isn't a punctuation? Or are there other punctuator prefixes that this should also cover?

danez: I looked through the acorn code and I'm pretty sure, that currently if `type.prefix` is true, the value is either `!` or `~`. I only added the value comparison, so that in future if acorn adds another prefix (which might not be a Punctator?) it is not immediately converted to Punctator.

Do you think it is safe to assume that prefix is always a Punctator?

danez:
It seems that my change also makes the testsuite hang on the XMLHttpRequest test. :(

lib/token-translator.js
nzakas: I think you can safely say if the first character of the token is not an identifier start, then it's a punctuator.

danez: I just checked again: `typeof`, `void` and `delete` are also [prefixes](https://github.com/ternjs/acorn/blob/9349eb4f4ec91da2b7236f2218e84a51bee22969/src/tokentype.js#L144). So a differentiation is necessary.

danez:
I finally fixed the tests. They were hanging because i forgot to update the library results and assert can't handle diffs that big.

nzakas:
This looks great, thanks! I'll do a release later today.

PR #277, opened 2016-05-26
Run travis tests on node 6

danez:


nzakas:
Thanks. Looks like there are some test failures with Node 6, can you look into that?

danez:
I had a brief look, but couldn't find what is going on. Seems that on node 6 the regexes are somehow evaluated somewhere. Might also be an issue in acorn.

danez:
Acorn generates different output depending on the system it is executed. The `value` of an Literal of type regex is `null` on node <=5 because `u` and `y` flag are not supported. But on node 6 these flags are working so acorn sets the regex as `value`.

I'm not sure if this behavior is ok for espree or it should be always null. If it is okay, then I'm not sure how to model the test to allow `null` or `RegExp`.

danez:
Also the exceptions that acorn throws are different, as on node <=5 it is validating the regexp by string matching, on node 6 it is just trying to execute the regex and if it fails it returns the exception that node/v8 produced.

nzakas:
Thanks for digging in. We probably just need to manually set `value` to always be `null` to account for this difference.

danez:
Yes I already tried that, and it would work, but not sure how to handle the different SyntaxError exceptions.

nzakas:
@danez can you open an issue for this? If you can also explain in the issue how you attempted to solve the problem and what trouble you had, that would help a lot.

mysticatea:
Closing as #315 fixed it.

PR #295, opened 2016-09-24
Fix: parsing error on `async` as property name.

mysticatea:
Fixes https://github.com/eslint/eslint/issues/7232

```js
const obj = {async: 1}
```

This is legal syntax, but espree 3.3.0 throws parsing error.


ljharb:
It'd also be good to ensure the following work:
```js
var async = 3;
var obj = {
  async() {}
};
class Foo {
  async() {}
}
```

mysticatea:
Good catch. I will add a test for the variable name.
There are tests for methods:

- [class A {async() { }};](https://github.com/eslint/espree/blob/6606f63a6c30ef5cd3405f0a4e46fd9f51f7fbbe/tests/fixtures/ecma-version/8/not-strict/async-named-class-method.src.js)
- [({async() { }});](https://github.com/eslint/espree/blob/6606f63a6c30ef5cd3405f0a4e46fd9f51f7fbbe/tests/fixtures/ecma-version/8/not-strict/async-named-object-method.src.js)

PR #301, opened 2016-10-20
Update: To support year in ecmaVersion number (fixes #300)

### LINKED ISSUE #300: Use official ECMA version numbers
The `ecmaVersion` key should also accept the official ECMA version numbers:

- 2015 instead of 6
- 2016 instead of 7
- …

gyandeeps:


kaicataldo:
Do we want to duplicate the normalizing logic in ESLint still? https://github.com/eslint/eslint/blob/cbbe420433726a5c7f38b132efac2945053966d5/lib/eslint.js#L427

gyandeeps:
I will work on that once I have this PR merged.

PR #305, opened 2016-11-06
Fix: classify `await` tokens as keywords (fixes #304)

### LINKED ISSUE #304: `await` token is incorrectly classified as an Identifier instead of a Keyword
```js
// espree 3.3.2

console.log(require('espree').parse('(async function() { await foo; })', {ecmaVersion: 8, tokens: true}).tokens);
```

Output:

```
[ Token { type: 'Punctuator', value: '(', start: 0, end: 1 },
  Token { type: 'Identifier', value: 'async', start: 1, end: 6 },
  Token { type: 'Keyword', value: 'function', start: 7, end: 15 },
  Token { type: 'Punctuator', value: '(', start: 15, end: 16 },
  Token { type: 'Punctuator', value: ')', start: 16, end: 17 },
  Token { type: 'Punctuator', value: '{', start: 18, end: 19 },
  Token { type: 'Identifier', value: 'await', start: 20, end: 25 },
  Token { type: 'Identifier', value: 'foo', start: 26, end: 29 },
  Token { type: 'Punctuator', value: ';', start: 29, end: 30 },
  Token { type: 'Punctuator', value: '}', start: 31, end: 32 },
  Token { type: 'Punctuator', value: ')', start: 32, end: 33 } ]
```

Note that `await` is classified as an `Identifier` instead of a `Keyword`.

not-an-aardvark:
Previously, `await` tokens were classified as `Identifier`s (see https://github.com/eslint/espree/issues/304). However, this is incorrect; `await` is a keyword similar to `yield`. This fixes the token translator to classify `await` as a keyword.

### DIFF
--- a/lib/token-translator.js
+++ b/lib/token-translator.js
@@ -110,7 +110,7 @@ TokenTranslator.prototype = {
                 token.type = Token.Keyword;
             }
 
-            if (extra.ecmaVersion > 5 && (token.value === "yield" || token.value === "let")) {
+            if (extra.ecmaVersion > 5 && (token.value === "yield" || token.value === "let" || token.value === "await")) {
                 token.type = Token.Keyword;
             }
 
--- a/tests/fixtures/ecma-version/6/modules/valid-await.result.js
+++ b/tests/fixtures/ecma-version/6/modules/valid-await.result.js
@@ -126,7 +126,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-arrow-func-parens.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-arrow-func-parens.result.js
@@ -309,7 +309,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-arrow-expression.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-arrow-expression.result.js
@@ -220,7 +220,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-arrow-param.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-arrow-param.result.js
@@ -366,7 +366,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-class-method-param.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-class-method-param.result.js
@@ -532,7 +532,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-destructured-default.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-destructured-default.result.js
@@ -455,7 +455,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-expression-class-method.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-expression-class-method.result.js
@@ -386,7 +386,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-function-param.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-function-param.result.js
@@ -455,7 +455,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-identifier-math.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-identifier-math.result.js
@@ -274,7 +274,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-inside-parens.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-inside-parens.result.js
@@ -344,7 +344,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-math.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-math.result.js
@@ -398,7 +398,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
@@ -452,7 +452,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
--- a/tests/fixtures/ecma-version/8/modules/async-await-named-object-method.result.js
+++ b/tests/fixtures/ecma-version/8/modules/async-await-named-object-method.result.js
@@ -188,7 +188,7 @@ module.exports = {
             ]
         },
         {
-            "type": "Identifier",
+            "type": "Keyword",
             "value": "await",
             "loc": {
                 "start": {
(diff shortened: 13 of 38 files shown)


not-an-aardvark:
That's a good point; this PR handles `await` the same way as `yield`. It looks like the existing implementation for `yield` has a similar bug:

```js
require('espree').parse('var yield = 1;', {tokens: true, ecmaVersion: 6}).tokens
```

```
[
  Token { type: 'Keyword', value: 'var', start: 0, end: 3 },
  Token { type: 'Keyword', value: 'yield', start: 4, end: 9 }, // <-- this should be Identifier
  Token { type: 'Punctuator', value: '=', start: 10, end: 11 },
  Token { type: 'Numeric', value: '1', start: 12, end: 13 }
]
```

not-an-aardvark:
Actually, this appears to be broken for other keywords as well:

```js
require('espree').parse('({try: 1})', {tokens:true})
```

```
[
  Token { type: 'Punctuator', value: '(', start: 0, end: 1 },
  Token { type: 'Punctuator', value: '{', start: 1, end: 2 },
  Token { type: 'Keyword', value: 'try', start: 2, end: 5 }, // <-- this should be Identifier
  Token { type: 'Punctuator', value: ':', start: 5, end: 6 },
  Token { type: 'Numeric', value: '1', start: 7, end: 8 },
  Token { type: 'Punctuator', value: '}', start: 8, end: 9 },
  Token { type: 'Punctuator', value: ')', start: 9, end: 10 }
]
```

I wonder if we should merge this and then correct things like contextual keywords in a separate issue.

nzakas:
@not-an-aardvark `try` is still a keyword, it's just a keyword used in a property context. That's the correct behavior.

I'm not entirely sure about `yield`. I know we had some issues open about whether `yield` should be a keyword or not, so I'd suggest searching through those for some context.

In any event, I'd suggest we keep this discussion focused on `await` for now.

not-an-aardvark:
Hmm, I think I'm misunderstanding the correct behavior then. My understanding was that token classification is context-based, i.e. something like `try` is only a Keyword if it's usedd in a Keyword position (i.e. a `TryStatement`), and something like `await` is only a Keyword if it's used in a Keyword position (i.e. in an `AwaitExpression`). Based on your last comment, it looks like I'm wrong; could you clarify the correct behavior?

mysticatea:
I would like to clear "what tokens are Keyword/Identifier?".

- **A:** Keyword tokens are the production [ReservedWord](https://tc39.github.io/ecma262/#prod-ReservedWord). Identifier tokens are the production [Identifier](https://tc39.github.io/ecma262/#prod-Identifier).
- **B:** Keyword tokens are the tokens which are controlling syntax as a keyword; those are hard coded in syntax. Identifier tokens are the production [IdentifierName](https://tc39.github.io/ecma262/#prod-IdentifierName).

And then:

- If it's **A**, all of `try`, `yield` (>=2015), and `await` (>=2017) are keywords always. [IdentifierReference](https://tc39.github.io/ecma262/#sec-identifiers)-like productions are handling some keywords as special.
- If it's **B**, properties names are Identifier tokens always. About both `yield` (>=2015) and `await` (>=2017) as variable names, I'm troubling since those are hard coded in syntax. But [IdentifierReference](https://tc39.github.io/ecma262/#sec-identifiers)-like productions are called Identifier, so I think those are Identifier tokens.

By the way, should this discuss in Acorn layer? :smiley:

mysticatea:
Incidentally, in my first implementation, `await` was a keyword. But it was removed from keywords in https://github.com/ternjs/acorn/commit/b12370f4e8d6b34a02d5d820c4ded77d67fa21e4.

not-an-aardvark:
Thanks for the explanation.

With `{sourceType: module}`, `await` a [FutureReservedWord](https://tc39.github.io/ecma262/#sec-future-reserved-words), but it is also a [Keyword](https://tc39.github.io/ecma262/#prod-Keyword). I wonder if this is a typo in the spec.

I think the correct behavior is:

```js
// sourceType: script

var await = 1; // Identifier (from IdentifierReference)
var yield = 1; // Identifier (from IdentifierReference) (https://tc39.github.io/ecma262/#sec-identifiers-static-semantics-early-errors)
var let = 1; // Identifier

({ await: 1 }); // Keyword?
({ yield: 1 }); // Keyword
({ let: 1 }); // Identifier

function foo() {
  'use strict';

  var await = 1; // Identifier
  var let = 1; // (SyntaxError)
  var yield = 1; // (SyntaxError)
  ({ await: 1 }); // Keyword?
  ({ yield: 1 }); // Keyword
  ({ let: 1 }); // Identifier (see note in https://tc39.github.io/ecma262/#prod-Keyword)
}

function* foo() {
  yield 5; // Keyword
  ({ yield: 1 }); // Keyword
  ({ await: 1 }); // Keyword?
}

async function bar() {
  await 5; // Keyword
  ({ yield: 1 }); // Keyword
  ({ await: 1 }); // Keyword
  ({ let: 1 }); // Identifier
}

```

```js
// sourceType: module

var await = 1; // (SyntaxError)

({ await: 1 }); // Keyword? FutureReservedWord?
```

mysticatea:
@not-an-aardvark Thank you for examples. Hmm, if my understanding is corrrect, it's:

- About `await`:
  - [IdentifierName] is Keyword tokens.
  - [IdentifierReference] is Identifier tokens.
  - [BindingIdentifier] is Identifier tokens.
  - [LabelIdentifier] is Identifier tokens.

I'm wondering that [IdentifierName] is not Keyword.
I think the correct behavior is **A** or **B** in my feeling. 

----

> - **A:** Keyword tokens are the production ReservedWord. Identifier tokens are the production Identifier.

**A** is based on the fact that the token types are decided in tokenizing phase (before parsing syntax phase). It does not know which the token is Keyword or Identifier in tokenizing phase, so it's handling tokens as keywords if the token matches a keyword.

```js
let await = 1;  // Keyword (VariableDeclaration → BindingIdentifier → `await`)
({await: 1});  // Keyword (LiteralPropertyName → IdentifierName)
async function baz() {
  await 5; // Keyword (AwaitExpression → `await`)
  ({ await: 1 }); // Keyword (LiteralPropertyName → IdentifierName)
}

// In ES modules
let await = 1;  // Keyword (VariableDeclaration → BindingIdentifier → `await`)
```

----

> - **B:** Keyword tokens are the tokens which are controlling syntax as a keyword; those are hard coded in syntax. Identifier tokens are the production IdentifierName.

On the other hand, **B** is gotten feedback the result of parsing syntax. The tokens are keywords if those are working as keywords accurately.

```js
let await = 1;  // Identifier (VariableDeclaration → BindingIdentifier → `await`)
({await: 1});  // Identifier (LiteralPropertyName → IdentifierName)
async function baz() {
  await 5; // Keyword (AwaitExpression → `await`)
  ({ await: 1 }); // Identifier (LiteralPropertyName → IdentifierName)
}

// In ES modules
let await = 1;  // Identifier (VariableDeclaration → BindingIdentifier → `await`)
```

not-an-aardvark:
I asked about this on the #tc39 IRC channel.

<details>
<summary>Logs</summary>

```
18:16:59 <not-an-aardvark> I have a question about parsing semantics. https://tc39.github.io/ecma262/#sec-keywords seems to imply that `await` can't be used as an identifier in any program, but that seems incorrect (`var await = 1` should be valid with the script parsing goal).
18:18:02 <not-an-aardvark> The section immediately following that indicates that `await` is a FutureReservedWord in modules -- is it supposed to only be a keyword in modules as well?
18:19:19 <not-an-aardvark> (A similar question about how to classify `await` tokens also came up in https://github.com/eslint/espree/pull/305#issuecomment-260532352 .)
18:22:32 <+bterlson> not-an-aardvark: await is always a keyword in modules, and contextually a keyword in scripts (whether `await` is parsed as an identifier or an expression depends on the presence of the Await grammar parameter)
18:24:03 <+bterlson> not-an-aardvark: you may find the notes in this clause helpful: https://tc39.github.io/ecma262/#sec-async-function-definitions
18:24:03 <+Bakkot> bterlson: `await` is (unconditionally) a Keyword, which means it is a ReservedWord, which means it can't be used as an identifier, no?
18:24:56 <+leobalter> not-an-aardvark, bterlson: https://tc39.github.io/ecma262/#sec-identifiers-static-semantics-early-errors wrt await + module
18:25:32 <+bterlson> Bakkot: maybe not Identifier but it's a separate alternative for BindingIdentifier at least
18:26:15 <+Bakkot> ah!
18:26:17 <+leobalter> I just opened an issue on test262 to complete the coverage for disallowed identifiers
18:26:47 <+Bakkot> yeah. not-an-aardvark, that's the answer to your question: it isn't an Identifier, but it can still be used as an IdentifierReference or BindingIdentifier: https://tc39.github.io/ecma262/#sec-identifiers
18:27:01 <+Bakkot> (in the absence of the Await flag, and outside of the Module goal symbol)
18:27:39 <+leobalter> I'm sorry if anyone is getting notifications for test262, not so sorry.
18:27:57 <+leobalter> I'm doing a big triage, and I intend to continue this triage soon
18:28:16 <+bterlson> Bakkot: and a LabelIdentifier
18:28:32 <not-an-aardvark> Thanks everyone for the help. I am still a bit confused because the three spec passages that have been linked all seem to contradict each other.
18:28:36 <not-an-aardvark> The first one says `await` is unconditionally a keyword, the second one says `await` is always a keyword only when the [Await] parameter is present and is an identifier otherwise, and the third one calls for an early error if `await` appears as an identifier with the [Await] parameter (although it seems like the second clause would imply that this shouldn't
18:28:36 <not-an-aardvark> be possible).
18:30:54 <+bterlson> An `await` token is unconditionally parsed as a Keyword, but that does not mean it is a conceptual keyword. Whether it behaves like you expect depends on the context the 2nd link is describing.
18:32:32 <not-an-aardvark> I see. Is it similar to `yield` in that regard?
18:32:42 <+bterlson> very similar to yield
18:33:12 <not-an-aardvark> The note in 11.6.2.1 makes a specific exception for `yield`, but it doesn't have an exception for `await` -- is that intended?
18:34:19 <+bterlson> that note should probably include mentions of await as well
18:35:39 <+bterlson> I have to think about the 12.1.1 errors more but at one point I was convinced they were needed
```
</details>

Conclusion:

* In modules, `await` is always a keyword (similar to other keywords like `try`).
* In scripts, `await` is a keyword when the `[Await]` grammar parameter is present (i.e. in async functions), and otherwise it's an identifier.

mysticatea:
Cool.
So, I think this should be handled in acorn layer since official statement is clear.

not-an-aardvark:
Agreed, I'll close this.

PR #308, opened 2016-11-28
Updated readme example to es6

amilajack:


nzakas:
Thanks for the submission. Can you explain why you think this is a good change? In general, we've been using ES5 as baseline for documentation as there are still a lot of people who aren't using ES6.

amilajack:
I see. Closing this to encourage more consistency then. I created the PR to encourage best practices, which, in my opinion, include using ES6.

PR #314, opened 2017-02-02
Build: Lock acorn to v4.0.4

kaicataldo:
See 
 - https://github.com/ternjs/acorn/issues/496
 - https://github.com/ternjs/acorn/issues/497

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   },
   "license": "BSD-2-Clause",
   "dependencies": {
-    "acorn": "^4.0.1",
+    "acorn": "4.0.4",
     "acorn-jsx": "^3.0.0"
   },
   "devDependencies": {


puzrin:
Note, 4.0.7 not hangs, but still only partially fixed another bug https://github.com/ternjs/acorn/issues/500#issuecomment-277128699. After forced updates i still see error report in eslint. It worth keep espree on 4.0.4 a bit more.

puzrin:
Update: 4.0.8 fixed things (see link above for details).

As author says, technically that fix is not perfect and should be improved. But 4.0.4 had the same problem. So now 4.0.8 should be not worse than 4.0.4.

gyandeeps:
Apparently `4.0.8` has some other issues: #316

puzrin:
Sorry for confusion, i meaned only very specific case like `[ (G = s), F = G ]`, which could be missed in tests.

nzakas:
Just a reminder: dependency changes should be "Upgrade:" or "Update" (in this case). This change is unrelated to the build.

PR #318, opened 2017-02-09
Add typings for typescript users.

rictic:


rictic:
Ah, I probably can't sign the CLA, will have to check in on that and get back to you.

PR #321, opened 2017-02-22
Add typings for typescript users.

rictic:


nzakas:
I'm sorry, we aren't going to be including TypeScript definitions in the package. We just don't have the resources to maintain this on an ongoing basis, and we don't want to get into a situation where we then need to support type definitions for multiple dialects.

rictic:
That's understandable. It's not much less convenient to have the types in DefinitelyTyped instead. Thanks for the quick review :)

JounQin:
@rictic So are you going to add is into DefinitelyTyped?

PR #323, opened 2017-03-19
Fix: raise error for trailing commas after rest properties (fixes #310)

### LINKED ISSUE #310: Trailing commas after rest properties should be a syntax error
```js
require('espree').parse(
    'const { foo, ...bar, } = baz;',
    { ecmaVersion: 6, ecmaFeatures: { experimentalObjectRestSpread: true } }
);
```

Expected behavior: SyntaxError

Actual behavior: Successful parse

---

ESLint has received a large number of invalid `comma-dangle` bug reports indirectly caused by confusion from this bug. https://github.com/eslint/eslint/issues/7422, https://github.com/eslint/eslint/issues/7463, https://github.com/eslint/eslint/issues/7502, https://github.com/eslint/eslint/issues/7614, https://github.com/eslint/eslint/issues/7857, https://github.com/eslint/eslint/issues/8273, https://github.com/eslint/eslint/issues/8289.

That said, I'm not sure we should fix this until we do a major release. Babel actually the [same issue](https://github.com/babel/babylon/pull/149) at one point and they [fixed it](https://github.com/babel/babylon/releases/tag/v6.11.3), but they had to [revert it](https://github.com/babel/babylon/releases/tag/v6.11.4) almost immediately afterwards due to a lot of ecosystem breakage.

Still, I thought I'd make a tracking issue for it since it's incorrect behavior.

not-an-aardvark:
This updates the rest property parser to raise an error if it encounters a trailing comma. The lack of an error here has caused a lot of confusion in eslint issues (see https://github.com/eslint/espree/issues/310 for a list of invalid bug reports).

This fixes incorrect behavior, but we should consider whether we want to treat it as a breaking change (see https://github.com/eslint/espree/issues/310).


nzakas:
If this change is due to a spec update, it should be breaking. If this was just a long-existing bug we didn't notice, then it's just a patch.

not-an-aardvark:
The spec has been consistent on this -- it's just a long-existing bug.

PR #327, opened 2017-03-28
Upgrade: acorn to 5.0.1

not-an-aardvark:
> ### Bug fixes
> Raise an error for duplicated lexical bindings.
> Fix spurious error when an assignement expression occurred after a spread expression.
> Accept regular expressions after `of` (in `for`/`of`), `yield` (in a generator), and braced arrow functions.
> Allow labels in front or `var` declarations, even in strict mode.
> ### Breaking changes
> Parse declarations following `export default` as declaration nodes, not expressions. This means that class and function declarations nodes can now have `null` as their `id`.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -194,13 +194,6 @@ function esprimaFinishNode(result) {
         }
     }
 
-    // Acorn currently uses expressions instead of declarations in default exports
-    if (result.type === "ExportDefaultDeclaration") {
-        if (/^(Class|Function)Expression$/.test(result.declaration.type)) {
-            result.declaration.type = result.declaration.type.replace("Expression", "Declaration");
-        }
-    }
-
     // Acorn uses undefined instead of null, which affects serialization
     if (result.type === "Literal" && result.value === undefined) {
         result.value = null;
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   },
   "license": "BSD-2-Clause",
   "dependencies": {
-    "acorn": "^4.0.11",
+    "acorn": "^5.0.1",
     "acorn-jsx": "^3.0.0"
   },
   "devDependencies": {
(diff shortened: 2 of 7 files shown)


not-an-aardvark:
> Can the breaking change affect us in any case?

Yes, our `FunctionDeclaration` ASTs will also sometimes have `id: null` as a result of this change.

I'm not sure whether this would be a bugfix for us under our semver policy.

mysticatea:
I think the breaking change is not breaking our use case.

We have been according to ESTree spec about `ExportDefaultDeclaration` since the first time. I think we can just remove the converting [espree.js#L198-L202](https://github.com/eslint/espree/blob/master/espree.js#L198-L202).

not-an-aardvark:
@mysticatea You're right, I hadn't realized we already convert the nodes. So this is not a breaking change for us.

However, when I tried removing the conversion, one of our tests failed due to https://github.com/ternjs/acorn/issues/528.

mysticatea:
Hmm? I think `export default (async function() {})` should be a `FunctionExpression` because of parentheses. If so, our test is wrong.

not-an-aardvark:
Well, in Acorn, `export default (function() {})` is a `FunctionDeclaration`. So I think it should be consistent either way.

mysticatea:
In the spec: https://tc39.github.io/ecma262/#prod-ExportDeclaration

If parentheses exist, it's the last case. So it's an `AssignmentExpression`.
If no parentheses, it's `export default HoistableDeclaration`, so it should be `FunctionDeclaration`.

This effects to the hoisting behavior.
The `foo` of `export default function foo() {}` is hoisting, but `export default (function foo() {});` is not hoisting.

not-an-aardvark:
@mysticatea I think you're right. But in that case, we have another bug: `export default (function() {})` is currently parsed as a `FunctionDeclaration` when it should be a `FunctionExpression`.

not-an-aardvark:
I noticed this introduces a new parsing error:

```js
(_ => 0) || (_ => 0)
```

This should parse successfully, but now espree fails to parse it.

PR #328, opened 2017-03-28
Upgrade: eslint to 2.13.1

not-an-aardvark:
This upgrades our `eslint` devDependency to the latest version that we can support while still supporting Node < 4. Previously, we were using `2.0.0-beta1`.


ilyavolodin:
Should we drop support for Node < 4 in Travis builds, since ESLint doesn't support it?

not-an-aardvark:
Oops, I forgot about that. I think it's still worth keeping the builds if we're going to support Node < 4.

Instead, I guess I'll update to ESLint 2.13.1.

PR #341, opened 2017-06-07
Fix: async shorthand properties (fixes #340)

### LINKED ISSUE #340: Using Object shorthand with a prop named "async" results in a parse error
Here's a reduced test case that generates the error: 

```javascript
/* global console */

const async = false;
const bar = 1;
const foo = { async, bar };

console.log(foo);
```
Error message is: error  Parsing error: Unexpected token ,

Eslint version is: v4.0.0-rc.0
espree: 3.4.3

Using ecmaversion 8.

Changing: `const foo = { async, bar };` to `const foo = { async: async, bar };` prevents the error.

Relates to #295

mysticatea:
This PR fixes #340.

I have sent PR to upstream (https://github.com/ternjs/acorn/pull/562), but we need fix espree as well because espree has overridden the method.


mysticatea:
I updated this PR according to the upstream review.

PR #342, opened 2017-06-19
Docs: Specify default values of options (fixes #325)

### LINKED ISSUE #325: Specify default values of options in the README
The Readme lists the available option of the `parse` function, but doesn't specify the default value for all of them.
From what I [gathered from](https://github.com/eslint/espree/blob/master/espree.js#L79) [the source](https://github.com/eslint/espree/blob/master/lib/features.js), it might make sense to update the example of the [usage section](https://github.com/eslint/espree#usage) to the following:
```js
var espree = require("espree");

var ast = espree.parse(code, {

    // attach range information to each node
    range: false,

    // attach line/column location information to each node
    loc: false,

    // create a top-level comments array containing all comments
    comment: false,

    // attach comments to the closest relevant node as leadingComments and
    // trailingComments
    attachComment: false,

    // create a top-level tokens array containing all tokens
    tokens: false,

    // set to 3, 5 (default), 6, 7, or 8 to specify the version of ECMAScript syntax you want to use. 
    // You can also set to 2015 (same as 6), 2016 (same as 7), or 2017 (same as 8) to use the year-based naming.
    ecmaVersion: 5,

    // specify which type of script you're parsing (script or module, default is script)
    sourceType: "script",

    // specify additional language features
    ecmaFeatures: {

        // enable JSX parsing
        jsx: false,

        // enable return in global scope
        globalReturn: false,

        // enable implied strict mode (if ecmaVersion >= 5)
        impliedStrict: false,

        // allow experimental object rest/spread
        experimentalObjectRestSpread: false
    }
});
```

Hirse:
@kaicataldo how does this look for you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ var ast = espree.parse(code, {
     // create a top-level tokens array containing all tokens
     tokens: false,
 
-    // set to 3, 5, 6, 7, or 8 to specify the version of ECMAScript syntax you want to use
-    // You can also set to 2015 (same as 6), 2016 (same as 7), or 2017 (same as 8) to use the year-based naming
+    // Set to 3, 5, 6, 7, or 8 to specify the version of ECMAScript syntax you want to use.
+    // You can also set to 2015 (same as 6), 2016 (same as 7), or 2017 (same as 8) to use the year-based naming.
     ecmaVersion: 5,
 
     // specify which type of script you're parsing ("script" or "module")
--- a/README.md
+++ b/README.md
@@ -28,45 +28,45 @@ There is a second argument to `parse()` that allows you to specify various optio
 ```javascript
 var espree = require("espree");
 
+// Optional second options argument with the following default settings
 var ast = espree.parse(code, {
 
     // attach range information to each node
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,45 +28,45 @@ There is a second argument to `parse()` that allows you to specify various optio
...
+    // attach comments to the closest relevant node as leadingComments and trailingComments
+    attachComment: false,
 
     // create a top-level tokens array containing all tokens
-    tokens: true,
+    tokens: false,
 
-    // set to 3, 5 (default), 6, 7, or 8 to specify the version of ECMAScript syntax you want to use. 
-    // You can also set to 2015 (same as 6), 2016 (same as 7), or 2017 (same as 8) to use the year-based naming.
+    // set to 3, 5, 6, 7, or 8 to specify the version of ECMAScript syntax you want to use
```
kaicataldo: For clarity, can we make both of these sentences (Captialize `set` and put periods at the end of each comment)

kaicataldo:
Thanks for contributing!

PR #343, opened 2017-06-19
Make template token objects adhere token object structure

### LINKED ISSUE #319: Template token objects differ from other token objects
While fixing an issue in `eslint-plugin-react` (https://github.com/yannickcr/eslint-plugin-react/issues/1061), I noticed that template tokens differ from other tokens. A typical token looks like this:

```
Token {
  type: 'String',
  value: '"bar"',
  start: 44,
  end: 49,
  loc:
   SourceLocation {
     start: Position { line: 4, column: 11 },
     end: Position { line: 4, column: 16 } },
  range: [ 44, 49 ] }
```

A template token looks like this:

```
{ type: 'Template',
  value: '`bar`',
  loc:
   { start: Position { line: 4, column: 11 },
     end: Position { line: 4, column: 16 } },
  range: [ 44, 49 ] }
```

I've not be able to figure out why templates are plain JavaScript objects and not of the class type `Token` (aside from the fact that the [template tokens are constructed differently in the token translator](https://github.com/eslint/espree/blob/58f75be6b89d8904b6366ed368045cb02c4a4e33/lib/token-translator.js#L43)). 

The more pressing issue, in my mind, is the omission of `start` and `end` properties at the top level of the template token object. The issue cited above was caused by passing `token.start` into eslint's `getNodeByRangeIndex` method. I fixed the issue by using `range[0]` (https://github.com/yannickcr/eslint-plugin-react/pull/1077), however I fear that other downstream projects are relying on tokens have a consistent object construction and always having `start` and `end`.

It seems like the quick-fix would be to copy the `range` values into `start` and `end` properties on the template tokens. If that would be sufficient for now, I'd be happy to PR that.

iancmyers:
While fixing an issue in `eslint-plugin-react` (https://github.com/yannickcr/eslint-plugin-react/issues/1061), I noticed that template tokens differ from other tokens. A typical token looks like this:

```
Token {
  type: 'String',
  value: '"bar"',
  start: 44,
  end: 49,
  loc:
   SourceLocation {
     start: Position { line: 4, column: 11 },
     end: Position { line: 4, column: 16 } },
  range: [ 44, 49 ] }
```

A template token looks like this:

```
{ type: 'Template',
  value: '`bar`',
  loc:
   { start: Position { line: 4, column: 11 },
     end: Position { line: 4, column: 16 } },
  range: [ 44, 49 ] }
```

I've not be able to figure out why templates are plain JavaScript objects and not of the class type `Token` (aside from the fact that the template tokens are constructed differently in the token translator.

This fix copies the `range` values into `start` and `end` properties on the template tokens to make them adhere to the same structure as other token objects.

Fixes #319


lib/token-translator.js
```diff
@@ -58,6 +58,8 @@ function convertTemplatePart(tokens, code) {
 
     if (firstToken.range) {
         token.range = [firstToken.range[0], lastTemplateToken.range[1]];
+        token.start = firstToken.range[0];
+        token.end = lastTemplateToken.range[1];
```
ljharb: nbd but maybe `token.range = [token.start, token.end]` after these lines would reduce some of the duplication?

iancmyers:
@ljharb I'd be happy to add a test, but I'd need to be pointed in the correct direction.

ljharb:
I think perhaps the [tokenize fixtures](https://github.com/eslint/espree/tree/master/tests/fixtures/tokenize) might need `start` and `end` added alongside `range` and `loc`, but I'm not 100% confident :-/

iancmyers:
@ljharb Looks like the tester explicitly strips out `start` and `end` and the root level of the token object: https://github.com/eslint/espree/blob/master/tests/lib/tester.js#L28

kaicataldo:
If we're not testing these values anywhere else in our tests it seems to me like we shouldn't hold this change up for that. If we decide to merge this as is, I can make a follow up issue to see if we can make this testable somehow.

As @nzakas mentioned in the original response, this seems like a pretty clear bug and an easy, low risk fix.

@eslint/eslint-team Does anyone know why the tester does this?

ljharb:
@kaicataldo totally; my comment wasn't meant as a blocker, more killing time/seeing if the PR can be improved while waiting for a response from eslint core ;-) it'd be great to merge it straightaway!

not-an-aardvark:
Thanks for contributing!

PR #356, opened 2017-10-09
Fix: Remove blockBindings from docs (fixes #307, fixes #339)

### LINKED ISSUE #307: README states ecmaFeatures.blockBindings is opt-in, but doesn't seem so; can't opt out
The README states:

> `let` and `const` declarations are no longer parsed by default. You must opt-in using `ecmaFeatures.blockBindings`.

However, it seems that `ecmaFeatures.blockBindings` is turned on by default, and that there is no way to opt out.

I am using Espree via ESLint. This is my `.eslintrc.yml`:

```yaml
extends:
  - eslint:recommended

parserOptions:
  sourceType: module
  ecmaFeatures:
    blockBindings: false
```

I have a `let` declaration in one of the files I am linting, but I do not see any errors.

At the end of the day, what I need is ES5 + import/export. Is there some other way to achieve this?

### LINKED ISSUE #339: blockBindings option has been removed
The README mentions a `blockBindings` option, which is absent from the code.

Note: I only intend to report the issue, not submit a PR.

Hirse:
This PR removes the mentioning of the `blockBindings` option from the *Differences from Espree 2.x* section in the README as it is no longer supported and the opt-in apparently not necessary anymore.

> ~let and const declarations are no longer parsed by default. You must opt-in using ecmaFeatures.blockBindings.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,6 @@ Espree is licensed under a permissive BSD 2-clause license.
 
 * The `tokenize()` method does not use `ecmaFeatures`. Any string will be tokenized completely based on ECMAScript 6 semantics.
 * Trailing whitespace no longer is counted as part of a node.
-* `let` and `const` declarations are no longer parsed by default. You must opt-in using `ecmaFeatures.blockBindings`.
 * The `esparse` and `esvalidate` binary scripts have been removed.
 * There is no `tolerant` option. We will investigate adding this back in the future.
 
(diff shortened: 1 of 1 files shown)


platinumazure:
Hi @Hirse, thanks for the pull request!

I believe one still has to opt in by specifying `ecmaVersion: 2015` (or greater). ES5 is still the default, so `let` and `const` will not be parsed with the default settings.

Hirse:
@platinumazure you are right. Based on some testing with the [AST explorer](https://astexplorer.net/) there are two ways to enable support for `let` and `const`: Using `ecmaVersion > 5` or setting `sourceType` to `module`.  I will update the docs accordingly.

PR #360, opened 2017-10-29
Chore: upgrade acorn@5.2 (fixes #358)

### LINKED ISSUE #358: ExpressionStatement#directive property
Before, ESTree spec added `ExpressionStatement#directive` property for directive prologues: https://github.com/estree/estree/compare/b3de58c9997504d6fba04b72f76e6dd1619ee4eb...1da8e603237144f44710360f8feb7a9977e905e0

We should follow the spec change.

I worked on acorn: https://github.com/ternjs/acorn/pull/593
I will update acorn after it's released.

mysticatea:
Fixes #358.
This PR upgrades acorn to `5.2.1`.

It has a feature and several bug fixes.

I fixed some tests because new acorn makes `ExpressionStatement#directive` property for directive prologues. This is by ESTree updating.

This PR is WIP since it doesn't finish the tests for `tests/fixtures/libraries/angular-1.2.5.js`.


mysticatea:
`chai.assert.equal(stringified_ast, result.json)` seems to not finish..

mysticatea:
OK, I fixed library tests. I believe the change of library tests is only `,"directive":"use strict"` parts.

PR #362, opened 2017-10-31
Breaking: drop supporting Node.js 0.x

Aladdin-ADD:


ilyavolodin:
We would have to do that in a major release.

PR #365, opened 2018-01-07
Upgrade: acorn v5.3.0

mysticatea:
This PR fixes some test failures that acorn v5.3.0 introduced.

- It added a new argument to some methods we are overriding: `toAssignable` and `checkPropClash`. It didn't seem to trigger crashes but caused to overlook some syntax errors.
- It improved the error message around `async`/`await`.


platinumazure:
@mysticatea Should this be an "Upgrade" commit?

mysticatea:
Good point. I agree.

PR #371, opened 2018-03-07
Breaking: require Node.js 6+, upgrade acorn-jsx@4.1.1 (fixes #345)

### LINKED ISSUE #345: Update acorn-jsx
As described [here](https://github.com/eslint/eslint/issues/6964#issuecomment-302288444), we should upgrade acorn-jsx to v4.0.1. Making an issue in case someone wants to take this on :)

Aladdin-ADD:


package.json
```diff
@@ -10,7 +10,7 @@
     "espree.js"
   ],
   "engines": {
-    "node": ">=0.10.0"
+    "node": ">=4.0.0"
```
not-an-aardvark: Since we're doing a breaking change, should we change this to `>=6.0.0` to be the same as ESLint v5?

tests/fixtures/ecma-version/9/regexp-lookbehind-assertion/valid-negative-1.result.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
                     0,
                     8
                 ],
-                "value": null,
+                "value": process.versions.node >= "9.0.0" ? new RegExp(String.raw`(?<!a)`) : null,
```
Aladdin-ADD: emm, this failed. seems Node.js 8 doesn't support these features, checking Node.js version seems not a good idea, thougths?

not-an-aardvark: For unicode regexes, the tests use the [`conditionalRegex`](https://github.com/eslint/espree/blob/0252144395e4a56ddaacb1a0b8eba693391e3d62/tests/lib/conditional-regex-value.js) helper. Would that work here?

It's not related to the issue, but I think using `>= "9.0.0"` will fail when Node 10.0.0 comes out, because it compares the version alphabetically.

tests/fixtures/ecma-version/9/regexp-lookbehind-assertion/valid-negative-1.result.js
```diff
@@ -1,3 +1,5 @@
+const condtionalRegex = require("../../../../lib/conditional-regex-value");
```
not-an-aardvark: Nitpick: I think the variable name is misspelled here (it is missing an "i" after "cond").

PR #381, opened 2018-06-04
Build: add node 10

Aladdin-ADD:


platinumazure:
I think we generally prefer to support node >= some version in most cases. Should we really remove node 7 here (even though it has long since hit end-of-life)?

mysticatea:
> should be the same?

Espree's support versions should be wide than ESLint's. If not, ESLint cannot support those versions.

PR #382, opened 2018-06-04
Chore: remove Object.assign polyfill

Aladdin-ADD:


espree.js
```diff
@@ -70,26 +70,7 @@ var DEFAULT_ECMA_VERSION = 5;
...
-
-            for (var keyIndex = 0; keyIndex < keys.length; keyIndex++) {
-                target[keys[keyIndex]] = arguments[argIndex][keys[keyIndex]];
-            }
-        }
-    }
-
-    return target;
-};
+var assign = Object.assign;
```
not-an-aardvark: Nitpick: it might be better to just use `Object.assign` directly where it's needed, rather than creating a variable called `assign`.

PR #383, opened 2018-06-09
Build: Adding rc release script to package.json

platinumazure:
Looking at the eslint-release code, it *looks* like we should be able to do an rc release of espree by supplying the rc prerelease label to the eslint-prerelease script. So I created a "rcrelease" script here to do just that.


platinumazure:
I goofed and used the wrong commit tag, so this should be merged with the PR title instead of the commit message. Sorry.

PR #387, opened 2018-07-06
Upgrade: eslint & eslint-config-eslint

Aladdin-ADD:
the pr upgrade these devdeps: 
+ "eslint": "^5.6.1",
+ "eslint-config-eslint": "^5.0.0",
+ "eslint-plugin-node": "^7.0.1",
+ "istanbul": "^0.4.5",


Aladdin-ADD:
it would be fine to add `eslint-plugin-node`. I added it in `eslint-config-eslint`, but it has not been released yet. Another choice is moving `espree` to eslint repo (monorepo). I created an issue (https://github.com/eslint/eslint/issues/10575).

mysticatea:
I feel built-in `assert` is nice enough :D

kaicataldo:
Do we want to wait on https://github.com/eslint/espree/pull/391 to land before doing this work? Looks like it might caused some merge conflicts.

Aladdin-ADD:
sure, I like to see `istanbul` => `nyc` (it has been done in #391 )

kaicataldo:
https://github.com/eslint/espree/pull/391 has been merged!

Aladdin-ADD:
@kaicataldo  I just rebased the newest changes, and thanks for the reminder! :)

Makefile.js
```diff
@@ -2,7 +2,7 @@
  * @fileoverview Build file
  * @author nzakas
  */
-/* global cp, echo, exit, find, mkdir, rm, target, test */
+/* global cp, echo, exit, mkdir, rm, target, test */
```
kaicataldo: Just to double check - is `find` no longer being used in this file?

Makefile.js
```diff
@@ -13,41 +13,23 @@
...
-// Helpers
-//------------------------------------------------------------------------------
-
-/**
- * Generates a function that matches files with a particular extension.
- * @param {string} extension The file extension (i.e. "js")
- * @returns {Function} The function to pass into a filter method.
- * @private
- */
-function fileType(extension) {
```
kaicataldo: Was this unused?

Makefile.js
```diff
@@ -112,21 +94,21 @@ target.browserify = function() {
     // 1. create temp and build directory
     if (!test("-d", TEMP_DIR)) {
         mkdir(TEMP_DIR);
-        mkdir(TEMP_DIR + "/lib");
+        mkdir(`${TEMP_DIR}/lib`);
```
kaicataldo: Looks like the `/` is unnecessary here. I wonder if we should just use `path.join()`?

Makefile.js
```diff
@@ -112,21 +94,21 @@ target.browserify = function() {
...
     }
 
     if (!test("-d", BUILD_DIR)) {
         mkdir(BUILD_DIR);
     }
 
     // 2. copy files into temp directory
-    cp("-r", "lib/*", TEMP_DIR + "/lib");
+    cp("-r", "lib/*", `${TEMP_DIR}/lib`);
```
kaicataldo: Looks like the `/` is unnecessary here. I wonder if we should just use `path.join()`?

Makefile.js
```diff
@@ -112,21 +94,21 @@ target.browserify = function() {
...
     cp("espree.js", TEMP_DIR);
     cp("package.json", TEMP_DIR);
 
 
     // 3. browserify the temp directory
-    nodeCLI.exec("browserify", TEMP_DIR + "espree.js", "-o", BUILD_DIR + "espree.js", "-s espree");
+    nodeCLI.exec("browserify", `${TEMP_DIR}espree.js`, "-o", `${BUILD_DIR}espree.js`, "-s espree");
```
kaicataldo: I wonder if we should just use `path.join()`?

Makefile.js
Aladdin-ADD: it was used before the change, but I replaced it with `lib/*.js`.

kaicataldo: Ah, I see - missed that. Thanks :+1:

Aladdin-ADD:
> are we trying to move towards using Node's assert module?

no, I don't think we have made the decision.

PR #388, opened 2018-07-25
visitor-keys add AwaitExpression

othree:
I found`lib/visitor-keys.js` don't have `AwaitExpression`


platinumazure:
@eslint/eslint-team Is this something we need for espree? I'm a bit confused, as we also have eslint-visitor-keys.

othree:
I found eslint-visitor-keys after I send this PR. And eslint-visitor-keys is updated.

Not sure what is the expected structure of eslint components.
But to reduce un-synchronize issue.
I think its better for espree to require eslint-visitor-keys.

platinumazure:
@othree Thanks for the information!

I agree- it would be nice if espree could use eslint-visitor-keys. But I'm not sure whether that's possible or if there is a specific reason for espree to have its own visitor keys. That's why I've pinged the team: I don't know the answer myself, but hopefully someone else on the team will. :smile:

not-an-aardvark:
Agreed, I think we should update this to start depending on `eslint-visitor-keys`.

othree:
@platinumazure I think this is just because eslint-visitor-keys is much newer than espree 😄

PR #389, opened 2018-09-13
VisitorKeys depend on eslint-visitor-keys

othree:
This PR is related to #388
Now the VisitorKeys is depend on https://github.com/eslint/eslint-visitor-keys


othree:
I didn't use VisitorKeys, and I don't think this util is widely used by people :)

ilyavolodin:
I don't think this require a major version change. The output should stay consistent, so an outside consumer should not see any difference (as far as I understand it).

mysticatea:
OK, I will merge this PR as an update. Thank you!

Issue #384, opened 2018-06-14
Option validation


platinumazure:
It could be useful to add option validation to espree and make sure that incoming options match a schema.

Are there any downsides? Do we want to allow variadic options? Should we make this validation itself opt-in?

(Inspired by eslint/eslint#10479.)

not-an-aardvark:
One potential downside is that ESLint occasionally passes new options to all parsers to allow for feature detection (e.g. [`eslintScopeManager`](https://eslint.org/docs/developer-guide/working-with-plugins#working-with-custom-parsers) and [`eslintVisitorKeys`](https://eslint.org/docs/developer-guide/working-with-plugins#working-with-custom-parsers)). If we throw an error for unrecognized option, then we might end up with unexpected crashes when ESLint adds a new option to pass to parsers.

platinumazure:
@not-an-aardvark Good call, I didn't think of that.

Maybe we could have a solution that allows some option wildcards (e.g., if the key starts with "eslint") but otherwise raise an error on unknown options?

Aladdin-ADD:
Take a step back, we can simply check the option types, not values. e.g.
```js
{
ecmaVersion: "2018"
}
```

we can give a warning like `ecmaVersion should be a number, but you are passing "2018", do you mean 2018?`

tapasweni-pathak:
:wave: @platinumazure @not-an-aardvark want to contribute to eslint codebase, working on a pr for this, do you think it would be good to implement a schema like [this](https://github.com/eslint/eslint/blob/master/conf/config-schema.js)? What specific options should be present in schema, what should be the error messages? great if you could add some more details in this issue.

kaicataldo:
@tapasweni-pathak Sorry we missed this. I think we need to discuss this change some more, but we'd love your contribution if you're still interested!

@not-an-aardvark @platinumazure Another option could be to only validate keys we know (and have documented), rather than introducing additional complexity around special key name patterns.

not-an-aardvark:
To clarify, do you mean we would validate the values of known options (e.g. making sure `ecmaVersion` is a number rather than a string) but we would continue to ignore unknown options? That sounds fine to me -- it seems like a clear improvement over what we do now.

kaicataldo:
That's correct.

tapasweni-pathak:
@kaicataldo @not-an-aardvark thanks for updating the ticket, picking this ticket, i'll ping in gitter for any other questions.

nzakas:
[...]

Issue #378, opened 2018-04-06
Readme/documentation updates around ESLint comment attachment strategy


platinumazure:
I see two locations in the README.md that potentially need to be changed due to ESLint's new comment attachment strategy (i.e., ESLint attaches comments itself and we no longer rely on espree to do it for us):

* Line 110: Is this statement still true? Did comment attachment *in espree* change in later versions than 2.x?
* Line 128: ESLint doesn't rely on espree's comment attachment anymore

(N.B. All line locations are based on commit 56c5a9c7fb01bd7ba338d16c1c63d89d5b8ee8b6, which is current as of the creation time for this issue.)

Also, should we consider adding a note that ESLint no longer uses comment attachment from espree and whether that has any implications on if we continue to support a comment attachment strategy in espree?

not-an-aardvark:
Regarding the first question, I think comment attachment in espree hasn't changed (we just toggled the option for it in ESLint).

Personally, I would be fine with removing comment attachment from espree given that ESLint no longer uses it.

kaicataldo:
I would love to remove comment attachment (and am happy to make a PR). I wonder if this is something we should discuss in our next TSC meeting, since it's a pretty big breaking change as far as the parser goes.

platinumazure:
@kaicataldo I would love to discuss it in a TSC meeting, but I'm wondering if we could create a new issue? This issue was originally about documentation updates that might be needed in espree's README. Thanks!

kaicataldo:
Done! https://github.com/eslint/espree/issues/394

PR #391, opened 2018-10-08
Upgrade: acorn 6, acorn-jsx 5, and istanbul

### LINKED ISSUE #390: Upgrade acorn to 6.x
[Release note of acorn@6.0.0](https://github.com/acornjs/acorn/blob/master/acorn/CHANGELOG.md#600-2018-09-14): It has been rewritten their plugin system drastically, so we need to rewrite Espree, too.

I will work on this.

mysticatea:
This PR upgrades some dependencies:

- `acorn` to 6.0.2 ... This major version has changed the plugin system drastically.
- `acorn-jsx` to 5.0.0 to use new plugin system.
- `istanbul` to `nyc`, the official successor of `istanbul`, to measure coverage for ES2015 syntax (I encountered some conversion errors).

This PR moves the `acorn` plugin part to `./lib/espree.js` from `./espree.js`. (...and I deleted some unused logics.) New plugin system is class-based, so I use symbols to store private stuff in order to vaoid confliction with acorn's private stuff.

Probably this upgrade will fix https://github.com/eslint/eslint/issues/10623.
Fixes #390.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -58,271 +58,38 @@
 
 "use strict";
 
-var astNodeTypes = require("./lib/ast-node-types"),
-    commentAttachment = require("./lib/comment-attachment"),
-    TokenTranslator = require("./lib/token-translator"),
-    acornJSX = require("acorn-jsx/inject"),
-    rawAcorn = require("acorn");
+const acorn = require("acorn");
+const jsx = require("acorn-jsx");
+const astNodeTypes = require("./lib/ast-node-types");
+const espree = require("./lib/espree");
 
+// To initialize lazily.
+const parsers = {
+    _regular: null,
+    _jsx: null,
 
-var acorn = acornJSX(rawAcorn);
-var DEFAULT_ECMA_VERSION = 5;
-var lookahead,
-    extra,
-    lastToken;
-
-/**
- * Resets the extra object to its default.
- * @returns {void}
- * @private
- */
-function resetExtra() {
-    extra = {
-        tokens: null,
-        range: false,
-        loc: false,
-        comment: false,
-        comments: [],
-        tolerant: false,
-        errors: [],
-        strict: false,
-        ecmaFeatures: {},
-        ecmaVersion: DEFAULT_ECMA_VERSION,
-        isModule: false
-    };
-}
-
-
-
-var tt = acorn.tokTypes,
-    getLineInfo = acorn.getLineInfo;
-
-// custom type for JSX attribute values
-tt.jsxAttrValueToken = {};
-
-/**
- * Normalize ECMAScript version from the initial config
- * @param  {number} ecmaVersion ECMAScript version from the initial config
- * @returns {number} normalized ECMAScript version
- */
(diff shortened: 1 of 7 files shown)


Aladdin-ADD:
> istanbul to nyc, the official successor of istanbul, to measure coverage for ES2015 syntax (I encountered some conversion errors).

I encountered it in #378, so I upgraded `istanbul@0.4.5`. Hopefully the pr can be merged soon. 😄

lib/espree.js
```diff
@@ -0,0 +1,314 @@
...
+
+                    if (options.attachComment === true) {
+                        commentAttachment.addComment(comment);
+                    }
+                }
+            }
+        }, code);
+
+        // TODO: remove global state.
+        commentAttachment.reset();
```
kaicataldo: At this point, since we don't use comment attachment in ESLint, I wonder if we should actually remove the comment attachment option in the future? I don't think it adds much value as of right now.

espree.js
```diff
@@ -58,271 +58,38 @@
...
-}
-
-acorn.plugins.espree = function(instance) {
-
-    instance.extend("finishNode", wrapFinishNode);
-
-    instance.extend("finishNodeAt", wrapFinishNode);
-
-    instance.extend("next", function(next) {
-        return /** @this acorn.Parser */ function() {
```
kaicataldo: Is this behavior already taken care of in the super class's method?

lib/espree.js
mysticatea: I agree. We can remove the comment attachment logic in the future.
But this PR focuses on upgrading acorn since it's a breaking change that removes `attarchComment` option.

lib/espree.js
```diff
@@ -0,0 +1,314 @@
...
+const ESPRIMA_FINISH_NODE = Symbol("espree's esprimaFinishNode");
+const tokTypes = Object.assign({}, acorn.tokTypes, jsx.tokTypes);
+
+/**
+ * Normalize ECMAScript version from the initial config
+ * @param {number} ecmaVersion ECMAScript version from the initial config
+ * @returns {number} normalized ECMAScript version
+ */
+function normalizeEcmaVersion(ecmaVersion) {
+    if (typeof ecmaVersion === "number") {
```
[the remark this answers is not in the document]
mysticatea: There is a related issue: #384.

espree.js
mysticatea: Yes. This override exists to address https://github.com/ternjs/acorn/issues/363 and it has been fixed.

Aladdin-ADD:
friendly ping @kaicataldo , does you comments addressed?

/cc. @mysticatea seems there's a conflict need to resolve.

lib/espree.js
kaicataldo: Yes, sorry I wasn't clear. Wasn't trying to suggest we change that in this PR!

mysticatea:
Thank you!

PR #392, opened 2018-10-17
Upgrade: eslint-release@1.0.0

not-an-aardvark:
(refs https://github.com/eslint/eslint/issues/10631)

This updates `eslint-release` to allow the package to still be published from the Jenkins server now that we have 2FA enabled on the bot account.

Some changes are also needed on the Jenkins server -- I plan to make those changes after this is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "chai": "^1.10.0",
     "eslint": "^2.13.1",
     "eslint-config-eslint": "^3.0.0",
-    "eslint-release": "^0.11.1",
+    "eslint-release": "^1.0.0",
     "esprima": "latest",
     "esprima-fb": "^8001.2001.0-dev-harmony-fb",
     "istanbul": "~0.2.6",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "chai": "^1.10.0",
     "eslint": "^2.13.1",
     "eslint-config-eslint": "^3.0.0",
-    "eslint-release": "^0.11.1",
+    "eslint-release": "^1.0.0",
```
platinumazure: Should this say 1.1? I suppose there's no harm since the current range doesn't intersect with this range, so hopefully people won't have 1.0 lying around locally.

not-an-aardvark: It could. My thought process is:

* 1.1.0 contains some nice UX improvements when manually running the script, but isn't strictly necessary for the release to work.
* Anyone installing after this change will get 1.1.0 anyway.
* It would be mildly annoying to update the commits, commit messages, and PR titles in all the different repos to switch to explicitly using 1.1.0.

Based on this I'm inclined to leave it as-is since there doesn't seem to be any harm in doing so, but let me know if I've misjudged the situation and specifying 1.1.0 explicitly would be beneficial.

platinumazure: Nope. Ship it!

PR #405, opened 2018-11-30
Breaking: remove attachComment

kaicataldo:
fixes https://github.com/eslint/espree/issues/394
fixes https://github.com/eslint/espree/issues/378

It looks like running `tools/update-tests.js` breaks tests right now (with what's currently in master). I just updated the tests for the libs, since they used `attachComment` (by commenting out sections of the script before running it). I'd like to see if I can fix those in a follow up PR, but would love to do it separately since this is already a large PR. Thoughts?


aminya:
For those who need this, I have registered the latest Espree with this feature:
https://www.npmjs.com/package/espree-attachcomment

```
npm install espree-attachcomment

```

PR #407, opened 2019-02-05
Upgrade: acorn@6.0.7

kaicataldo:
acorn@6.0.7 is now [fully ECMA-262-compliant](https://twitter.com/MarijnJH/status/1092357649067372544)!

There are two types of test changes included:
 * Error message changes in Acorn
 * [Acorn now checks that exported bindings are defined](https://github.com/acornjs/acorn/commit/f8c9a6797592a2a0f0250526b22b808eada2f78e), which means that we have to update a few tests that only have a named export declaration without first defining the exported binding.

I had trouble getting our existing `tools/update-tests.js` script to work, so I ended up just running Espree over the code snippets and updating them manually. I'll try to find some time in the future to look at the scripts and see if I can get them to work - I seem to remember that being an issue last time I tried to update tests too.


tests/fixtures/ecma-version/9/async-generator/invalid-yield-in-nested-function-4.result.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
-    "index": 35,
+    "index": 29,
     "lineNumber": 1,
-    "column": 36,
-    "message": "Unexpected token a"
-};
\ No newline at end of file
+    "column": 30,
+    "message": "The keyword 'yield' is reserved"
```
kaicataldo: 🎉

tests/fixtures/ecma-version/6/modules/export-named-as-default.result.js
```diff
@@ -1,87 +1,179 @@
 module.exports = {
     "type": "Program",
+    "start": 0,
```
kaicataldo: I'm noticing these are new after updating these tests. It looks like Acorn has been adding the "start" and "end" properties to the `Program` node for a while.

PR #408, opened 2019-02-16
Pass the sourceFile parser option through to acorn

papandreou:
I'd like to use espree in a project that has to generate source maps, and for that I need the `loc` objects to include the url/file name of the source. Luckily `acorn` already supports this feature, so it's just a matter of letting the `sourceFile` parser option through.

Esprima calls this option `source` so it might be worth it to mitigate that difference as well.

### DIFF
--- a/lib/espree.js
+++ b/lib/espree.js
@@ -99,6 +99,7 @@ module.exports = () => Parser => class Espree extends Parser {
             sourceType: isModule ? "module" : "script",
             ranges: options.range === true,
             locations: options.loc === true,
+            sourceFile: options.sourceFile,
 
             // Truthy value is true for backward compatibility.
             allowReturnOutsideFunction: Boolean(ecmaFeatures.globalReturn),
--- a/tests/lib/parse.js
+++ b/tests/lib/parse.js
@@ -81,5 +81,14 @@ describe("parse()", () => {
             espree.parse("foo", Object.freeze({ ecmaFeatures: Object.freeze({}) }));
         });
 
+        it("should pass sourceFile through", () => {
+            const ast = espree.parse("var foo = bar;", {
+                loc: true,
+                sourceFile: "the/source.js"
+            });
+
+            assert.strictEqual(ast.loc.source, "the/source.js");
+        });
+
     });
 });


platinumazure:
Seems reasonable to me, but I would be interested in hearing other opinions from the team. :+1:

not-an-aardvark:
Could you elaborate on how this affects the generated ASTs? You mentioned "I need the `loc` objects to include the url/file name of the source" -- does this change all of the `loc` objects in the AST somehow?

papandreou:
@not-an-aardvark, the `loc` objects gain a `source` property alongside `start` and `end`:

```js
> ast = require('espree').parse('foo = bar;', {loc: true, sourceFile: 'the/source.js'});
Node {
  type: 'Program',
  start: 0,
  end: 10,
  loc:
   SourceLocation {
     start: Position { line: 1, column: 0 },
     end: Position { line: 1, column: 10 },
     source: 'the/source.js' },
  body:
   [ Node {
       type: 'ExpressionStatement',
       start: 0,
       end: 10,
       loc: [SourceLocation],
       expression: [Node] } ],
  sourceType: 'script' }
```

The `source` property is a de facto standard for keeping track of where each ast node originated, and will be picked up by `escodegen` et al. when generating source maps:

```js
> require('escodegen').generate(ast, {sourceMap: true})
'{"version":3,"sources":["the/source.js"],"names":["foo","bar"],"mappings":"AAAAA,GAAA,GAAMC,GAAN"}'
```

mysticatea:
Why every node, rather than `Program` node, requires `loc.source` property?

nzakas:
My understanding is that Acorn already supports this option, so the request here is just that we support the same option rather than where the information should be represented in the AST.

I'm not a fan of adding Acorn-specific stuff into Espree, because there is no guarantee that we will continue to use Acorn as the basis for Espree in the future.

My questions then, is, is this feature specific to Acorn? Are there any other ESTree-compatible parsers that support this?

papandreou:
The estree spec actually mentions the `source` property in the docs for the `SourceLocation` interface: https://github.com/estree/estree/blob/master/es5.md#node-objects

As I mentioned, esprima supports the same feature, but calls the option `source` instead of acorn's `sourceFile`:

```js
> require('esprima').parse('foo = bar;', {loc: true, source: 'the/source.js'});
Script {
  type: 'Program',
  body:
   [ ExpressionStatement {
       type: 'ExpressionStatement',
       expression: [AssignmentExpression],
       loc: [Object] } ],
  sourceType: 'script',
  loc:
   { start: { line: 1, column: 0 },
     end: { line: 1, column: 10 },
     source: 'the/source.js' } }
```

babylon has the same feature, but calls the option `sourceFilename` and the loc property `filename` 🙄

papandreou:
> Why every node, rather than `Program` node, requires `loc.source` property?

When bundling or combining JavaScript files in other ways, the resulting code can contain bits and pieces from various sources. Keeping track of that at the ast node level is good enough to generate a source map.

not-an-aardvark:
I'm not strongly opposed to this, but it feels like this is a feature that the user of `espree` could handle themselves by traversing the AST afterwards. The `start` and `end` location properties are information that can only be reasonably created by the parser, but the `source` location property is information that the user already has anyway, so I'm not convinced it's worthwhile to provide it as redundant metadata.

papandreou:
Shouldn't I think of espree as a general purpose JavaScript parser?

not-an-aardvark:
I think so. I don't think adding redundant info to AST nodes should be a requirement for being considered a general purpose JS parser.

papandreou:
@not-an-aardvark, this is a very standard JS parser feature, attaching it after the fact would be pretty inconvenient. The feedback I'm getting here tells me that espree is mostly a utility for eslint, and that I should just use acorn directly.

PR #412, opened 2019-03-14
Breaking: validate parser options (fixes #384)

### LINKED ISSUE #384: Option validation

aladdin-add:
it will throw an error if any of the conditions is true:

+ ecmaVersion is invalid
+ sourceType is invalid
+ ecmaVersion < 6 & sourceType: "module"

is there some more conditions need to be covered?

### DIFF
--- a/lib/espree.js
+++ b/lib/espree.js
@@ -63,7 +63,7 @@ function normalizeSourceType(sourceType = "script") {
  * @throws {Error} throw an error if found invalid option.
  * @returns {Object} normalized options
  */
-function validate(options) {
+function normalizeOptions(options) {
     const ecmaVersion = normalizeEcmaVersion(options.ecmaVersion);
     const sourceType = normalizeSourceType(options.sourceType);
 
@@ -114,7 +114,7 @@ module.exports = () => Parser => class Espree extends Parser {
         if (typeof code !== "string" && !(code instanceof String)) {
             code = String(code);
         }
-        const options = validate(opts);
+        const options = normalizeOptions(opts);
 
         const ecmaFeatures = options.ecmaFeatures || {};
         const ecmaVersion = normalizeEcmaVersion(options.ecmaVersion);
--- a/lib/espree.js
+++ b/lib/espree.js
@@ -10,12 +10,14 @@ const STATE = Symbol("espree's internal state");
 const ESPRIMA_FINISH_NODE = Symbol("espree's esprimaFinishNode");
 const tokTypes = Object.assign({}, acorn.tokTypes, jsx.tokTypes);
 
+
 /**
  * Normalize ECMAScript version from the initial config
  * @param {number} ecmaVersion ECMAScript version from the initial config
+ * @throws {Error} throws an error if the ecmaVersion is invalid.
  * @returns {number} normalized ECMAScript version
  */
-function normalizeEcmaVersion(ecmaVersion) {
+function normalizeEcmaVersion(ecmaVersion = DEFAULT_ECMA_VERSION) {
     if (typeof ecmaVersion === "number") {
         let version = ecmaVersion;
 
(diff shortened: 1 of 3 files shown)


lib/espree.js
```diff
@@ -36,11 +38,39 @@ function normalizeEcmaVersion(ecmaVersion) {
...
+ * @param {Object} options the parser options to normalize
+ * @throws {Error} throw an error if found invalid option.
+ * @returns {Object} normalized options
+ */
+function normalizeOptions(options) {
+    const ecmaVersion = normalizeEcmaVersion(options.ecmaVersion);
+    const sourceType = normalizeSourceType(options.sourceType);
+
+    if (sourceType === "module" && ecmaVersion < 6) {
+        throw new Error("sourceType 'module' is not supported when ecmaVersion < 6!");
```
platinumazure: I don't think we should use an exclamation point here-- it's unnecessary, and could potentially confuse people.

```suggestion
        throw new Error("sourceType 'module' is not supported when ecmaVersion < 6");
```

tests/lib/ecma-version.js
```diff
@@ -141,6 +141,30 @@ describe("ecmaVersion", () => {
             }, /Invalid ecmaVersion/);
         });
 
+        it("Should throw error using invalid year", () => {
```
platinumazure: This test case has the same name as the previous test case. Is that intentional?

One possible suggestion (feel free to disregard):

```suggestion
        it("Should throw error when non-numeric year is provided", () => {
```

lib/espree.js
```diff
@@ -36,11 +38,41 @@ function normalizeEcmaVersion(ecmaVersion) {
                 return version;
 
             default:
```
not-an-aardvark:
```suggestion
            // no default
```

lib/espree.js
```diff
@@ -36,11 +38,41 @@ function normalizeEcmaVersion(ecmaVersion) {
...
+ * @returns {Object} normalized options
+ */
+function normalizeOptions(options) {
+    const ecmaVersion = normalizeEcmaVersion(options.ecmaVersion);
+    const sourceType = normalizeSourceType(options.sourceType);
+    const ranges = options.range === true;
+    const locations = options.loc === true;
+
+    if (sourceType === "module" && ecmaVersion < 6) {
+        throw new Error("sourceType 'module' is not supported when ecmaVersion < 6.");
```
not-an-aardvark: It seems like this error will often be shown to some ESLint users. Should we make it more instructive about how to fix the problem?

```suggestion
        throw new Error("sourceType 'module' is not supported when ecmaVersion < 2015. Consider adding `{ ecmaVersion: 2015 }` to the parser options.");
```

not-an-aardvark:
Does this also fix https://github.com/eslint/eslint/issues/9687, or are more changes in ESLint needed to fix that?

aladdin-add:
> Does this also fix eslint/eslint#9687, or are more changes in ESLint needed to fix that?

we need to upgrade espree, since this will be in a major release. not sure some more, like adding a test?

simonbrunel:
So does that mean ESLint shouldn't be used in projects targeting `ecmaVersion < 2015`, while not using any transpiler but for which sources rely on ES6 import/export bundler (e.g. rollup). I don't think ESLint should enforce the `ecmaVersion >= 2015`  when `sourceType: module` since this should be an acceptable use case:

```javascript
module.exports = {
    parserOptions: {
        ecmaVersion: 5,       // disable ES6 features (const, let, etc.)
        sourceType: module    // allow ESM import/export syntax
    }
}
```

not-an-aardvark:
@simonbrunel To be clear, ESLint already does not support that case. The config you showed currently has the effect of setting `ecmaVersion: 2015` despite the `ecmaVersion: 5` value provided in the config. This change just makes the error more obvious to avoid setting any misleading expectations.

simonbrunel:
@not-an-aardvark yes I know, but these kind of projects exist and I don't think ESLint should exclude them.

simonbrunel:
What I mean is instead of setting `ecmaVersion` to `2015` automatically or trigger an error if it doesn't match when `sourceType: module`, it may be better to make a distinction between the scope of these 2 options and allow projects that use ESM import/export to disable other ES6 features.

aladdin-add:
you can use [eslint-plugin-es](https://www.npmjs.com/package/eslint-plugin-es).

simonbrunel:
@aladdin-add thanks, I will give it a try!

PR #415, opened 2019-07-13
Update: support bigint and dynamic import

mysticatea:
This PR adds supports of bigint and dynamic import.

- It enables bigint and dynamic import if `ecmaVersion` option was `2020` or `11`.
- It represents bigint literal as `Literal` node that has `bigint` property. The `bigint` property has the source code text of the bigint literal without the suffix `n`.
- It represents dynamic import as `ImportExpression` node. The `source` property is the importing source as similar to `ImportDeclaration` node, but the `source` property can be an arbitrary expression node.


tests/lib/tester.js
```diff
@@ -29,6 +29,11 @@ function getRaw(ast) {
             return undefined; // eslint-disable-line no-undefined
         }
 
+        // JSON cannot handle BigInt.
+        if (typeof value === "bigint") {
+            return null;
```
mysticatea: The tests depend on JSON heavily, but JSON doesn't support BigInt values. So tests handle `Literal#value` as null. I think that this is not a big problem because it's null if runtime doesn't support bigint natively.

platinumazure:
@mysticatea I've tried to push a few changes to consume the new acorn and eslint-visitor-keys, and then fixed a failing test. Please review the latest commits on this branch and let me know if there are issues. (Thinking about it now, I should have made a PR targeting this branch so you could review using the PR review tools. Sorry about that.)

mysticatea:
@platinumazure Thank you so much!

mysticatea:
LGTM.

.travis.yml
```diff
@@ -8,5 +7,10 @@ node_js:
     - 10
     - 11
     - 12
+
+script:
+    - if [ $TRAVIS_NODE_VERSION -ge 8 ]; then node Makefile.js lint; fi
+    - node Makefile.js test
```
aladdin-add: out of curiosity: this was added since eslint v6 don't support node <8?

we may consider dropping node <8, as it has been EOL.

platinumazure: @aladdin-add I assume that was the case (but hopefully @mysticatea can confirm). I think it makes sense to release a minor version of espree with this functionality if we can, and then we could do a major release to drop Node <8.

PR #416, opened 2019-07-25
Fix: a failing test with acorn 6.2.0

mysticatea:
This PR fixes a failing test case that is caused by the acorn update.

`for (const a = 0 of b) ;` should be a syntax error.


platinumazure:
I think I agree that we should upgrade the acorn dependency here- otherwise, merging this will cause tests to start failing for anyone with acorn@<6.2.0.

@mysticatea Any concerns?

platinumazure:
@mysticatea Is this still needed now that #415 is merged? I'm thinking maybe this is no longer needed.

mysticatea:
No. I'm sorry that I have lost this issue...

PR #419, opened 2019-07-30
Fix: tokenize the latest right curly brace (fixes #403)

### LINKED ISSUE #403: espree.tokenize doesn't tokenize } if it's last token
Espree doesn't tokenize `}`

```js
espree.tokenize('{}') // 1st
espree.tokenize('}') // 2nd
espree.tokenize('{}?') // 3rd
espree.tokenize('{\n}}') // 4th
```
output is:
```js
// 1st
[ Token { type: 'Punctuator', value: '{', start: 0, end: 1 } ]
// 2nd
[]
// 3rd
[ Token { type: 'Punctuator', value: '{', start: 0, end: 1 },
  Token { type: 'Punctuator', value: '}', start: 1, end: 2 },
  Token { type: 'Punctuator', value: '?', start: 2, end: 3 } ]
// 4th
[ Token { type: 'Punctuator', value: '{', start: 0, end: 1 },
  Token { type: 'Punctuator', value: '}', start: 2, end: 3 } ]
```

finico:
There is a logic in `TokenTranslator::onToken` about the latest curly brace
https://github.com/eslint/espree/blob/6ebc21947166399a0b4918d4a1beb9d610650336/lib/token-translator.js#L196-L204
But it executes only at first `eof` token (it seems strange that the first token is `eof`, but it comes from Acorn's init state)
I've added a call of `next` for the `TokenTranslator::onToken` to be executed with the last `eof` token

And I've noticed two failed tests, but they've already fixed at #416

### DIFF
--- a/lib/espree.js
+++ b/lib/espree.js
@@ -175,6 +175,9 @@ module.exports = () => Parser => class Espree extends Parser {
             this.next();
         } while (this.type !== tokTypes.eof);
 
+        // Consume the final eof token
+        this.next();
+
         const extra = this[STATE];
         const tokens = extra.tokens;
 
--- a/tests/lib/tokenize.js
+++ b/tests/lib/tokenize.js
@@ -199,6 +199,10 @@ describe("tokenize()", () => {
         espree.tokenize("foo", Object.freeze({ ecmaFeatures: Object.freeze({}) }));
     });
 
+    /**
+     * Make sure we tokenize closing curly brace in a block statement at end of file
+     * @see https://github.com/eslint/espree/issues/403 for more information
+     */
     it("should produce tokens when } is the last token", () => {
         const tokens = espree.tokenize("{}");
 
--- a/tests/lib/tokenize.js
+++ b/tests/lib/tokenize.js
@@ -199,4 +199,16 @@ describe("tokenize()", () => {
         espree.tokenize("foo", Object.freeze({ ecmaFeatures: Object.freeze({}) }));
     });
 
+    it("should produce tokens when } is the last token", () => {
+        const tokens = espree.tokenize("{}");
+
+        assert.deepStrictEqual(
+            tester.getRaw(tokens),
+            [
+                { type: "Punctuator", value: "{" },
+                { type: "Punctuator", value: "}" }
+            ]
+        );
+    });
+
 });


kaicataldo:
Hi, thanks for contributing. It looks like there are some tests failing - do you mind fixing those so that we can review this?

finico:
Hi @kaicataldo , thank you for quick response.
As I mentioned in the description, these tests were fixed in 416th pull request.
We can wait till it will be merged, then I will rebase my branch on upstream

lib/espree.js
```diff
@@ -174,6 +174,9 @@ module.exports = () => Parser => class Espree extends Parser {
             this.next();
         } while (this.type !== tokTypes.eof);
 
+        // it's a call for the latest `eof` token
```
kaicataldo: I think the word "latest" is confusing here because it looks like `this.next()` will continue returning muliple `eof` tokens. I wonder if something like the following would be clearer:

```suggestion
        // Consume the final eof token
```

platinumazure:
Hi @finico, would you mind merging/rebasing upstream changes please? Thanks!

finico:
@platinumazure @kaicataldo done

tests/lib/tokenize.js
```diff
@@ -199,4 +199,16 @@ describe("tokenize()", () => {
         espree.tokenize("foo", Object.freeze({ ecmaFeatures: Object.freeze({}) }));
     });
 
+    it("should produce tokens when } is the last token", () => {
```
kaicataldo: Do you mind adding a comment with a link to the relevant issue here so that future contributors have some context as to why this test exists?

finico: I've added, but I'm not sure about the description

tests/lib/tokenize.js
```diff
@@ -199,4 +199,20 @@ describe("tokenize()", () => {
         espree.tokenize("foo", Object.freeze({ ecmaFeatures: Object.freeze({}) }));
     });
 
+    /**
+     * Make sure we tokenize closing curly brace in a block statement
```
platinumazure: Based on what you understand of the issue, would this be an improvement?

```suggestion
     * Make sure we tokenize closing curly brace in a block statement at end of file
```

tests/lib/tokenize.js
kaicataldo: Sorry, could have been clearer. This is fine, but I also think a comment that just says `// https://github.com/eslint/espree/issues/403` would suffice!

tests/lib/tokenize.js
finico: I think so

tests/lib/tokenize.js
finico: Sorry, if you would suggest me what to add, I will add

tests/lib/tokenize.js
finico: added

tests/lib/tokenize.js
kaicataldo: This is fine! Thanks for adding the comment.

PR #420, opened 2019-08-13
Upgrade: eslint-visitor-keys@^1.1.0

platinumazure:


kaicataldo:
Actually, it looks like https://github.com/eslint/espree/pull/415 includes this already.

kaicataldo:
Sorry for the confusion here - I approved this and then noticed the above.

platinumazure:
Yes, you're right, I updated that PR myself even. Closing...

PR #421, opened 2019-08-15
Update: improve error messaging when validating ecmaVersion

kaicataldo:
Inspired by https://github.com/eslint/eslint/issues/12101. This PR enhances the error messaging when validating `ecmaVersion` to distinguish between when the the value provided is the incorrect data type versus when it's not a valid numeric value.

I labeled this an enhancement since it's not technically a bug fix, but please let me know if you disagree!

### DIFF
--- a/lib/espree.js
+++ b/lib/espree.js
@@ -18,8 +18,10 @@ const tokTypes = Object.assign({}, acorn.tokTypes, jsx.tokTypes);
  * @returns {number} normalized ECMAScript version
  */
 function normalizeEcmaVersion(ecmaVersion = DEFAULT_ECMA_VERSION) {
-    if (typeof ecmaVersion !== "number") {
-        throw new Error("ecmaVersion must be a number.");
+    const ecmaVersionType = typeof ecmaVersion;
+
+    if (ecmaVersionType !== "number") {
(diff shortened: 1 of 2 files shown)


kaicataldo:
Whoops, fixed the commit message 😅

lib/espree.js
```diff
@@ -18,28 +18,30 @@ const tokTypes = Object.assign({}, acorn.tokTypes, jsx.tokTypes);
  * @returns {number} normalized ECMAScript version
  */
 function normalizeEcmaVersion(ecmaVersion = DEFAULT_ECMA_VERSION) {
-    if (typeof ecmaVersion === "number") {
-        let version = ecmaVersion;
+    if (typeof ecmaVersion !== "number") {
+        throw new Error("ecmaVersion must be a number.");
```
ljharb:
```suggestion
        throw new Error("ecmaVersion must be a number, got `" + typeof ecmaVersion + "`.");
```

kaicataldo: Good idea!

ljharb:
(fwiw using `typeof x === 'string'` inline actually hits more optimizations in engines, making it much cheaper than storing the value in a variable and using it twice; nbd tho)

kaicataldo:
Cool, useful to know :+1:

PR #429, opened 2020-02-12
Breaking: drop Node v8 support

kaicataldo:
Now that Node v8 is no longer being supported (and we're dropping support in ESLint core), it seems prudent to also drop support in Espree.

Questions:
Have I missed anything?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,11 @@
 language: node_js
 node_js:
-    - 6
-    - 7
-    - 8
-    - 9
     - 10
-    - 11
     - 12
+    - 13
 
 script:
-    - if [ $TRAVIS_NODE_VERSION -ge 8 ]; then node Makefile.js lint; fi
+    - node Makefile.js lint
     - node Makefile.js test
 
 after_success:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,15 +1,11 @@
 language: node_js
 node_js:
-    - 6
-    - 7
-    - 8
-    - 9
     - 10
```
btmills: Do you think it's worth testing specifically on our minimum supported version similar to the [GitHub Actions tests in eslint/eslint](https://github.com/eslint/eslint/blob/1ee6b6388305a8671c8d4c3cf30c2dbf18a1ff7e/.github/workflows/ci.yml#L29)?

```suggestion
    - "10.12.0"
    - 10
```

anikethsaha:
@kaicataldo  I noticed few TODOs which were blocking cause of node < 8 . like [`this`], so should those be land in this PR or separate !?

kaicataldo:
Do you mind clarifying what you mean? Did you mean to link something?

anikethsaha:
sorry , forgot to add the link. here is the link 
https://github.com/eslint/espree/blob/master/lib/espree.js#L70

nzakas:
@anikethsaha we can handle that separately as it doesn’t block this PR.

PR #430, opened 2020-02-12
Update: add latestEcmaVersion & supportedEcmaVersions

kaicataldo:
Inspired by [this discussion](https://github.com/eslint/eslint/pull/12879#discussion_r376182196), this is a proposal to add `latestEcmaVersion()` & `supportedEcmaVersions()` to the public API.

Benefits:
1. We can query the latest `ecmaVersion` from the parser for rules (as is mentioned in the comment listed above).
1. The supported `ecmaVersion`s in our demo have gotten out of sync with ESLint core. This would allow us to calculate this programmatically and not have to manually update it.
1. This is a fairly minimal API that shouldn't require much maintenance. We will most likely adjust this code once a year.

Cons:
1. This increases the surface area of our API and will have to be maintained.

Open questions:
1. If we use this in core, are we expecting 3rd party parsers to also implement this? I would say no (and that we should always fall back to Espree in the event it doesn't exist on a 3rd party parser). I believe this is still better than it being hardcoded in core.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -82,15 +82,15 @@ function normalizeOptions(options) {
  * Get the latest ECMAScript version supported by Espree.
  * @returns {number} The latest ECMAScript version.
  */
-function latestEcmaVersion() {
-    return Math.max(...SUPPORTED_VERSIONS);
+function getLatestEcmaVersion() {
+    return SUPPORTED_VERSIONS[SUPPORTED_VERSIONS.length - 1];
 }
 
 /**
  * Get the list of ECMAScript versions supported by Espree.
  * @returns {number[]} An array containing the supported ECMAScript versions.
  */
-function supportedEcmaVersions() {
+function getSupportedEcmaVersions() {
     return [...SUPPORTED_VERSIONS];
 }
 
@@ -100,6 +100,6 @@ function supportedEcmaVersions() {
 
 module.exports = {
     normalizeOptions,
-    latestEcmaVersion,
-    supportedEcmaVersions
+    getLatestEcmaVersion,
+    getSupportedEcmaVersions
 };
--- a/lib/options.js
+++ b/lib/options.js
@@ -1,0 +1,105 @@
+/**
+ * @fileoverview A collection of methods for processing Espree's options.
+ * @author Kai Cataldo
+ */
+
+"use strict";
+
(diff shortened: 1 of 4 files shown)


lib/options.js
```diff
@@ -0,0 +1,105 @@
...
+        throw new Error("sourceType 'module' is not supported when ecmaVersion < 2015. Consider adding `{ ecmaVersion: 2015 }` to the parser options.");
+    }
+    return Object.assign({}, options, { ecmaVersion, sourceType, ranges, locations });
+}
+
+/**
+ * Get the latest ECMAScript version supported by Espree.
+ * @returns {number} The latest ECMAScript version.
+ */
+function latestEcmaVersion() {
```
nzakas: I’d suggest calling this getLatestEcmaVersion to indicate a value is being retrieved (verb as opposed to noun).

lib/options.js
```diff
@@ -0,0 +1,105 @@
...
+    return Math.max(...SUPPORTED_VERSIONS);
+}
+
+/**
+ * Get the list of ECMAScript versions supported by Espree.
+ * @returns {number[]} An array containing the supported ECMAScript versions.
+ */
+function supportedEcmaVersions() {
```
nzakas: Similarly, getSupportedEcmaVersions.

lib/options.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    return Object.assign({}, options, { ecmaVersion, sourceType, ranges, locations });
+}
+
+/**
+ * Get the latest ECMAScript version supported by Espree.
+ * @returns {number} The latest ECMAScript version.
+ */
+function latestEcmaVersion() {
+    return Math.max(...SUPPORTED_VERSIONS);
```
nzakas: You can avoid a calculation here by using SUPPORTED_VERSIONS[SUPPORTED_VERSIONS.length - 1]

lib/options.js
kaicataldo: Good call! I was initially thinking of it being more of a static property (like `version`), but that makes more sense with this iteration :+1:

kaicataldo:
Thanks for the review! Updated.

kaicataldo:
Ah, I like that! That’s more in line with my initial design and I think is a nicer API. Will update.

PR #432, opened 2020-02-28
Update: update acorn-jsx and fix failing test

kaicataldo:
`acorn-jsx` released v5.2.0, which includes the changes from [this PR](https://github.com/acornjs/acorn-jsx/pull/108). Since we have the same test in this repo, tests are failing in `master`. Updating the package.json and tests will ensure that tests pass again.

I've labeled this as an "Update" because I believe it should be treated as a semver-minor bug fix, since it is a bug fix that will lead to more errors.

More details can be found [here](https://github.com/acornjs/acorn-jsx/issues/106).

### DIFF
--- a/tests/fixtures/ecma-features/jsx/embedded-tags.src.js
+++ b/tests/fixtures/ecma-features/jsx/embedded-tags.src.js
@@ -1,1 +1,1 @@
-<LeftRight left=<a /> right=<b>monkeys /> gorillas</b> />;
+<LeftRight left=<a /> right=<b>monkeys /&gt; gorillas</b> />;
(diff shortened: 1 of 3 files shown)


tests/fixtures/ecma-features/jsx/embedded-tags.src.js
```diff
@@ -1 +1 @@
-<LeftRight left=<a /> right=<b>monkeys /> gorillas</b> />;
```
kaicataldo: This is a syntax error in the JSX spec. Please see https://github.com/acornjs/acorn-jsx/issues/106 for more details.

PR #439, opened 2020-04-09
Chore: added tests for options normalize

anikethsaha:
added tests for `lib/options.js`

### DIFF
--- a/tests/lib/options.js
+++ b/tests/lib/options.js
@@ -16,7 +16,7 @@ const assert = require("assert"),
 // Tests
 //------------------------------------------------------------------------------
 
-describe("options unit testing", () => {
+describe("normalizeOptions", () => {
     it("should throw error for sourceType module and ecmaVersion < 6", () => {
         const option = {
             sourceType: "module",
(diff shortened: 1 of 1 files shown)


anikethsaha:
@kaicataldo I agree I think there are 1-2 tests doing similar but in option testing, I am doing unit test for normalize method instead of running the whole parser. yes, it's not to strong as under the hood, its the same. 
I would like to know what you suggest !.  cause i think too that this `options`  testing is not so important atm as its already being done under the hood. 

let me know

kaicataldo:
I'm not against unit testing this module, but I do think there's a lot more value in doing a more integration-style test at the parser level.

tests/lib/options.js
```diff
@@ -0,0 +1,60 @@
...
+//------------------------------------------------------------------------------
+
+const assert = require("assert"),
+    { normalizeOptions } = require("../../lib/options");
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+describe("options unit testing", () => {
```
kaicataldo, line 19: Can we change this to the following, since we're only testing the one public method of this module at the moment?

```suggestion
describe("normalizeOptions", () => {
```

kaicataldo:
Thanks for contributing!

PR #440, opened 2020-04-09
Chore: added fixlint in Makefile.js

anikethsaha:
added a fixlint command and script in order to fix the autofixable linting issue

### DIFF
--- a/Makefile.js
+++ b/Makefile.js
@@ -42,62 +42,31 @@ target.all = function() {
 };
 
 target.lint = function() {
+
+    const flag = process.argv[process.argv.length - 1];
     let errors = 0,
         lastReturn;
 
     echo("Validating Makefile.js");
-    lastReturn = nodeCLI.exec("eslint", MAKEFILE);
+    lastReturn = nodeCLI.exec("eslint", MAKEFILE, flag);
     if (lastReturn.code !== 0) {
         errors++;
     }
 
     echo("Validating configuration files");
-    lastReturn = nodeCLI.exec("eslint", CONFIG_FILES);
+    lastReturn = nodeCLI.exec("eslint", CONFIG_FILES, flag);
     if (lastReturn.code !== 0) {
         errors++;
     }
 
     echo("Validating JavaScript files");
-    lastReturn = nodeCLI.exec("eslint", JS_FILES);
+    lastReturn = nodeCLI.exec("eslint", JS_FILES, flag);
     if (lastReturn.code !== 0) {
         errors++;
     }
 
     echo("Validating JavaScript test files");
-    lastReturn = nodeCLI.exec("eslint", TEST_FILES);
-    if (lastReturn.code !== 0) {
-        errors++;
-    }
-
-    if (errors) {
-        exit(1);
-    }
-};
-
-target.fixlint = function() {
-    let errors = 0,
-        lastReturn;
-
-    echo("Validating and fixing Makefile.js");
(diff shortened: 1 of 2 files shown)


Makefile.js
```diff
@@ -74,13 +74,53 @@ target.lint = function() {
     }
 };
 
+target.fixlint = function() {
```
nzakas, line 77: Is there a way we can avoid duplicating all of this logic from the lint target? It seems easy to forget to update one when updating the other.

anikethsaha: i can see two approaches in this to keep it consistent. 
- using `process.argv` to get the `fix` flag and run it in `target.lint` method itself.
- extract the logic from both in a separate function and simply pass a `fix` parament to the function call for `fixlint`

IMO, I would go with the first approach of using `process.argv`

Makefile.js
```diff
@@ -42,29 +42,30 @@ target.all = function() {
 };
 
 target.lint = function() {
+    const flag = process.argv[process.argv.length - 1] === "--fix" ? "--fix" : "";
```
kaicataldo, line 45: I think this is certainly better than duplicating all the logic, but it does feel kind of brittle. Could we just all arguments through?

anikethsaha: > Could we just all arguments through?

I didn't get this.  
you meant passing all arguments from `process.argv` ?

kaicataldo: That's right :+1:

anikethsaha: done 👍

Makefile.js
```diff
@@ -46,25 +46,25 @@ target.lint = function() {
         lastReturn;
 
     echo("Validating Makefile.js");
-    lastReturn = nodeCLI.exec("eslint", MAKEFILE);
+    lastReturn = nodeCLI.exec("eslint", MAKEFILE, process.argv.slice(3));
```
kaicataldo, line 49: Assigning `process.argv.slice(3)` to a variable that is reused (`cliFlags`, maybe?) will make it less error prone in the future (since we'll only have to change it in one place). Do you mind updating that? Otherwise, this LGTM!

kaicataldo:
@anikethsaha Do you mind resolving the merge conflicts?

Makefile.js
```diff
@@ -32,7 +32,10 @@ const NODE_MODULES = "./node_modules/",
     MAKEFILE = "Makefile.js",
     CONFIG_FILES = ".eslintrc.js",
     JS_FILES = "\"lib/**/*.js\" \"espree.js\"",
-    TEST_FILES = "tests/lib/**/*.js";
+    TEST_FILES = "tests/lib/**/*.js",
+
+    // FLAGS
+    CLIFLAGS = process.argv.slice(3);
```
kaicataldo, line 38: Looking at the documentation, I’m not sure `nodeCLI.exec()` accepts an array as an argument. Should this be the following?

```suggestion
    CLIFLAGS = process.argv.slice(3).join(“ “);
```

Alternatively, we could use the spread operator in the function calls below:

```js
lastReturn = nodeCLI.exec("eslint", MAKEFILE, ...CLIFLAGS);
```

anikethsaha: I am a bit confused as in [this line](https://github.com/nzakas/shelljs-nodecli/blob/master/lib/shelljs-nodecli.js#L76) it does say that the args are converted an array but [here](https://github.com/nzakas/shelljs-nodecli/blob/master/lib/shelljs-nodecli.js#L90) it is looping the argument.

I think its better to use the spread operator.

PR #441, opened 2020-04-19
Update: support `??` operator, `import.meta`, and `export * as ns`

mysticatea:
This is WIP.

This PR upgrades acorn to support new syntax:

- ~~[ ] Optional Chaining (waiting for upstreams)~~ Separated to other PR.
- [X] Nullish Coalescing
- [X] `export * as ns from "source"`
- [X] `import.meta`


package.json
```diff
@@ -18,9 +18,9 @@
   },
   "license": "BSD-2-Clause",
   "dependencies": {
-    "acorn": "^7.1.1",
+    "acorn": "^7.2.0",
     "acorn-jsx": "^5.2.0",
-    "eslint-visitor-keys": "^1.1.0"
+    "eslint-visitor-keys": "github:eslint/eslint-visitor-keys"
```
btmills, line 23: Flagging this just so we don't forget to update it to the new version once we release `eslint-visitor-keys`.

PR #446, opened 2020-06-11
Update: support optional chaining

mysticatea:
This PR adds support for optional chaining.

- Upgrade `acorn`
- Upgrade `eslint-visitor-keys`
- Modify token translation logic for the new punctuator token `?.`.


package.json
```diff
@@ -18,9 +18,9 @@
   },
   "license": "BSD-2-Clause",
   "dependencies": {
-    "acorn": "^7.2.0",
+    "acorn": "^7.3.1",
     "acorn-jsx": "^5.2.0",
-    "eslint-visitor-keys": "^1.2.0"
+    "eslint-visitor-keys": "github:eslint/eslint-visitor-keys#add-chain-expression"
```
mysticatea, line 23: We need to update it to a new version of `eslint-visitor-keys` after we released it.

nzakas: I just did a release of `eslint-visitor-keys`. Should be safe to update this now.

mysticatea: Thank you. I updated this line.

SilencerWeb:
Hi! Is there any way to know at least an estimated date when a new version with this will be released so ESLint would be able to parse optional chaining? 🙂

aladdin-add:
we do a release in every 2 weeks, so it will be in the next release(7.4.0). please follow the release issue: https://github.com/eslint/eslint/issues/13443

SilencerWeb:
@aladdin-add, thanks a lot!

PR #449, opened 2020-08-10
Chore: move to GitHub Actions

kaicataldo:
This PR switches us over from Travis to GitHub Actions. I removed the `npm run coveralls` step, since that's no longer in our `package.json`. I also made the Node versions matrix we're using to match the main ESLint repo's, since it's vital that this project works on those platforms. Let me know if that doesn't make sense for whatever reason!


.travis.yml
```diff
@@ -1,13 +0,0 @@
...
-    - 10
-    - 12
-    - 13
-
-script:
-    - node Makefile.js lint
-    - node Makefile.js test
-
-after_success:
-    - npm run coveralls
```
nzakas, line 13: Are we giving up on coveralls?

kaicataldo: I'm definitely not against it using coveralls. I removed this because this `npm` script doesn't currently exist and we don't have [node-coveralls](https://github.com/nickmerwin/node-coveralls) in our dev dependencies. But maybe I'm missing something? I looked through the Travis docs to make sure there wasn't some magic happening here, and I couldn't find anything.

nzakas: Ah ok. Just wanted to make sure we weren't missing anything.

PR #463, opened 2020-12-11
Chore: Add test for valid non-string input

stonegray:
Test that buffers and strings produce equal output to resolve uncovered line.

### DIFF
--- a/tests/lib/parse.js
+++ b/tests/lib/parse.js
@@ -96,5 +96,14 @@ describe("parse()", () => {
             espree.parse("foo", Object.freeze({ ecmaFeatures: Object.freeze({}) }));
         });
 
+        it("Cast valid non-string input", () => {
+            const str = "var foo = bar;";
+
+            assert.deepStrictEqual(
+                espree.parse(Buffer.from(str)),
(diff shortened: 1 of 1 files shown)


btmills:
Wow, recursive `git blame` traces that uncovered string coercion code all the way back to https://github.com/jquery/esprima/commit/8e3d670ec2bdb3a9c6c5d6f51893c781ba328f3e in 2012. Quite the history lesson! Given this behavior has been around longer than `espree` itself, I think we should test it as-is. Before accepting it unilaterally, I'll give the rest of the team a chance to chime in. I could be persuaded that we'd rather remove the coercion and say we officially only accept string inputs. That'd be a breaking change to be safe, but we have a major version coming up soon anyway.

nzakas:
I don’t think we need to make any changes, I’m fine with just adding the test.

anikethsaha:
Yeah , adding this test seems fine. additionally if we want to inform users, then we can add a note in docs saying its recommended to use string as inputs.

btmills (approved):
Seems we're in agreement. Thanks @stonegray! LGTM.

Issue #464, opened 2020-12-12
Proposal: Expose Acorn options


stonegray:
Acorn has a number of options that are quite useful, such as `allowReturnOutsideFunction`, when dealing with partial snippets of code, such as individual statements, or the contents of a single function. Currently, only some of these options are copied into the Acorn constructor, making a number of them inaccessible.

I propose addition of an `acornOptions`  object that can be used to directly set Acorn options, to allow both existing and future Acorn features to be controlled. 

This change would be non-breaking, and would not affect code coverage.

I've implemented this proposal here: 

* Commit: https://github.com/stonegray/espree/commit/b5e051fce76558decbecaa4233a1e172d9e6c173
* Branch: https://github.com/stonegray/espree/tree/acorn-options

nzakas:
Thanks for the idea. We purposely don’t expose Acorn options because the fact Espree currently uses Acorn doesn’t mean it always will, and we don’t want a leaky abstraction that forces us to implement every one of Acorn’s options. In fact, Espree started out life as it’s own parser and only later adopted Acorn transparently.

PR #465, opened 2020-12-12
Update: Expose Acorn options (#464)

stonegray:
Implementation of proposal #464 to expose Acorn options.

- Allow setting options for underlying Acorn parser using new acornOptions property of espree options object.
- Resolve TODO for repeated key copy as all targed versions of Node support the spread operator.


stonegray:
Unsure if this should be tagged `New` or `Update`, used update for now since it only exposes existing functionality, a " backwards-compatible enhancement". I can change this if needed.

PR #489, opened 2021-04-29
Chore: Refactor `TemplateElement` range fix

fisker:
Fix ranges/locations/start/end in one step, and use `TemplateElement.tail` instead of slicing the original input.


nzakas (approved):
LGTM

mdjermanovic (approved):
LGTM, thanks for contributing!

PR #509, opened 2021-07-09
Chore: Remove obsolete lib/visitor-keys.js

matz3:
PR #389 deleted "lib/visitor-keys.js" and replaced all usages with the
eslint-visitor-keys package.
PR #387 added lib/visitor-keys.js back again for no apparent reason.

It looks like #387 has been started before #389, so it needed to
integrate those changes and by mistake the file has been added again.


aladdin-add (approved):
good catch, thanks! LGTM, but I'd like one more reviewer.

aladdin-add:
can you sign our cla, so we can merge it?

matz3:
Signed 👍🏻

snitin315 (approved):
LGTM

mdjermanovic (approved):
LGTM, thanks!

That module is definitely unused now. Also, its content is old.

Since we have the `exports` field in package.json it can't be accessed, so it feels safe to remove the file.

matz3:
Thanks!
Yeah we also got confused by the different content, which is why I tried to understand why the file still exists.

PR #515, opened 2021-09-16
Chore: add class static blocks tests

aladdin-add:
acorn has been upgraded in ee1d3eca310cab1b1cf1563294434977d26358d4, this commit simply added some tests. The tests were copied from test262, refs: https://github.com/tc39/test262/commit/afe217b318df5f197e64b30a8b5a4b391c777359

### DIFF
--- a/tests/fixtures/ecma-version/13/class-static-blocks/statements-class-static-init-expr-new-target.result.js
+++ b/tests/fixtures/ecma-version/13/class-static-blocks/statements-class-static-init-expr-new-target.result.js
@@ -1,0 +1,890 @@
+export default {
+    "type": "Program",
+    "loc": {
+        "start": {
+            "line": 15,
+            "column": 0
+        },
+        "end": {
+            "line": 23,
+            "column": 35
+        }
+    },
+    "range": [
+        527,
+        633
+    ],
+    "body": [
+        {
+            "type": "VariableDeclaration",
+            "loc": {
+                "start": {
+                    "line": 15,
+                    "column": 0
+                },
+                "end": {
+                    "line": 15,
+                    "column": 17
+                }
+            },
+            "range": [
+                527,
+                544
+            ],
+            "declarations": [
+                {
+                    "type": "VariableDeclarator",
+                    "loc": {
+                        "start": {
+                            "line": 15,
+                            "column": 4
+                        },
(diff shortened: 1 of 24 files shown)


tests/fixtures/ecma-version/13/class-static-blocks/expressions-arrow-function-static-init-await-binding.src.js
```diff
@@ -0,0 +1,6 @@
+
```
aladdin-add, line 1: I'll try to add the test262 url here. thoughts?

nzakas: Can you explain what you mean by this? I'm not understanding.

aladdin-add: I was wondering to add a link to https://github.com/tc39/test262/commit/afe217b318df5f197e64b30a8b5a4b391c777359, as it was copied from there.

nzakas: You’d need to add not just a link but the full copyright and license.

aladdin-add: great, will do!

aladdin-add:
it might be the non-error tests diff was not displayed by default. 😄 

image

nzakas:
Ah ok. I don’t think we need so many tests that result in errors. We don’t need all of the test-262 cases because we are just wrapping what Acorn does. For other new features we’ve just done a handful of tests.

aladdin-add:
just removed some invalid tests.

nzakas (commented):
Do we need all of the tests that just flag “await” as an error? That seems like more than is necessary for this.

aladdin-add:
> Do we need all of the tests that just flag “await” as an error? That seems like more than is necessary for this.

some of these were thrown in "script" mode, should be resolved now. 🚀

nzakas (approved):
LGTM. Thanks!

PR #534, opened 2022-02-02
chore: switch from nyc to c8 for ESM coverage

brettz9:
Coverage has been broken since the ESM upgrade (we could add a threshold check to prevent such regressions, though noting that we're just shy of 100% due to https://github.com/eslint/espree/blob/main/lib/espree.js#L132 not being covered).


bmish:
What's the difference between the two?

brettz9:
> What's the difference between the two?

Mostly just the fact that nyc doesn't yet support ESM, and our project switched to ESM. In most other ways, it is just like `nyc`.

bmish:
Makes sense. I saw this change made here too: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1686

It looks like `.gitignore` and `.nycrc` need to be updated too.

brettz9:
> Makes sense. I saw this change made here too: [sindresorhus/eslint-plugin-unicorn#1686](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1686)
> 
> It looks like `.gitignore` and `.nycrc` need to be updated too.

No need for `nycrc` as c8 supports these files too. And no additional folders need supporting for coverage. (And I think we should keep `.nyc_output` in `.gitignore` in case reverting back to nyc as we could do (or might need to do for some reason or other) in the future once the new version of nyc may support ESM.)

bmish (approved):

brettz9:
Thanks for the review!

PR #551, opened 2022-06-19
ci: remove node v17

amareshsm:
** Description: **

Removed node v17 since it has reached EOL.


mdjermanovic (commented):
For the same reasons as https://github.com/eslint/eslint/pull/16027#pullrequestreview-1011558016, I think we should keep this for now.

btmills:
Thanks again @amareshsm, but holding off on this until the next major version.

Issue #593, opened 2024-01-06
espree improvements as an acorn plugin extension [RFC]


martypdx:
I'm building an eslint custom parser extension and wanted to leverage the espree parser as an acorn extension to take advantage of the esprima compatible translation that already exist for tokens and comments (and template differences).

Running into some issues with the structure of `espree`:

Extension function not exposed

The [acorn factory extension method](https://github.com/eslint/espree/blob/main/lib/espree.js#L55) is not currently exposed as an package export. Would be great to be able to import espreeAcornPlugin as a composable acorn plugin.

Changed option key for locations and ranges

Minor issue in that the keys for locations and ranges options differs (unnecessarily?) from acorn options. New options are fine but would be good to allow `locations` and `ranges` as input options.

~~Token and Context Types~~

~~This is a broader acorn plugin ecosystem issue, but the token types in espree are hard coded for only core `acorn` and `acorn-jsx`. I followed the `acorn-jsx` example and added a custom namespace as well, but that approach doesn't really work in a composable plugin system.  Not sure what the answer is. Have some hacks in mind, will update with more info when I try them.~~

EDIT: Not sure what the answer is for this. I ended up injecting new tokens into the base Parser. Opened [this issue](https://github.com/acornjs/acorn/issues/1271) on acorn.

nzakas:
We don’t expose Acorn related functionality because using Acorn id an implementation that could potentially change in the future. We’ve discussed this before. 

https://github.com/eslint/espree/issues/542#issuecomment-1094469829

PR #594, opened 2024-01-08
fix: Espree usage as acorn plugin

martypdx:
Changes to support #593:

1. Add export of plugin function to main module which allows use as:
  ```js
  import { espreeAcornPlugin as espree } from 'espree';
  
  const AzothEspreeParser = AcornParser.extend(azoth(), espree());
  ```

2. Respect acorn settings alias for `locations` and `ranges`. Still requires setting new, additive options `tokens` and `comment` but at least minimizes head-scratching over why standard option keys aren't working.

### DIFF
--- a/espree.js
+++ b/espree.js
@@ -175,3 +175,5 @@ export const Syntax = (function() {
 export const latestEcmaVersion = getLatestEcmaVersion();
 
 export const supportedEcmaVersions = getSupportedEcmaVersions();
+
+export { espree as espreeAcornPlugin };
--- a/lib/options.js
+++ b/lib/options.js
@@ -92,8 +92,8 @@ function normalizeSourceType(sourceType = "script") {
 export function normalizeOptions(options) {
     const ecmaVersion = normalizeEcmaVersion(options.ecmaVersion);
     const sourceType = normalizeSourceType(options.sourceType);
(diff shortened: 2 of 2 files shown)


nzakas:
We don’t expose Acorn related functionality because using Acorn id an implementation that could potentially change in the future. We’ve discussed this before. 

https://github.com/eslint/espree/issues/542#issuecomment-1094469829

martypdx:
Hi @nzakas, thanks for your reply and consideration of this request. 

So the whole "acorn is a detail" is a head-scratcher given there is nothing else in espree besides implementation details of how to adapt an acorn/acorn-jsx parser i/o to work with expected esprima format for eslint. Just about anyone creating a custom parser for eslint is going to have to deal with these set of issues. 🤷 

While disappointing, these are things that can be worked around. The examples are helpful, appreciation to all the contributors 

As you are working on the eslint@next ecosystem, please consider maintaining support for custom parsers and other forms of technology innovation through syntax and ast structure. It's much harder today than it was a decade ago to create the support tooling around extensions and customization. Typescript doesn't support syntax enhancements (though you can use VSCode language service to work around that). 

Eslint and ecosystem provide an important accessible entry for open technology innovation. I think it's critical to make sure that doesn't disappear.

PR #633, opened 2024-09-26
fix: Remove node:assert dependency

nzakas:
#### Prerequisites checklist

-   [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What changes did you make? (Give an overview)

Removed `node:assert` as a dependency in `eslint-scope`, replacing it with a custom `assert` function.

#### Related Issues

Fixes #625 
Closes #626

PR #634, opened 2024-09-27
chore: remove assert from rollup external array

amareshsm:
#### Prerequisites checklist

As part of #633, the `assert` dependency was removed, so there’s no need to include it as an external dependency in the Rollup config.


nzakas (approved):
LGTM. Thanks!

PR #664, opened 2025-06-12
chore: replace `npm-run-all2` with a native script

### LINKED ISSUE #660: Change Request: Replace `npm-run-all2` with a native script to reduce the dependency footprint

- [x] `espree`
- [ ] `eslint-scope`
- [ ] `eslint-visitor-keys`

In this issue, I would like to suggest replacing the `npm-run-all2` package with a native script to reduce the dependency footprint.

Currently, the `npm-run-all2` package is only used in the `espree` package as a devDependency, and it is only used when running the test scripts `test:cjs` and `test:esm`.

The `npm-run-all` script can be simply replaced with a native script:

```diff
- "npm-run-all -s test:*"
+ "npm run test:cjs && npm run test:esm"
```

Also, `eslint-visitor-keys` follows this strategy:
https://github.com/eslint/js/blob/main/packages/eslint-visitor-keys/package.json#L51

### Participation

- [x] I am willing to submit a pull request for this change.

### Additional comments

_No response_

lumirlumir:
#### Prerequisites checklist

Hello,

In this PR, I've replaced `npm-run-all2` with a native script.

This PR resolves #660.

While reducing devDependencies may not be critical in package development, I believe it's a good practice for the following reasons:

1. Performance - Faster install times and a smaller `node_modules`.
2. Maintenance - Fewer tools to update and manage.
3. Cleaner CI/CD - Leaner builds and more reliable pipelines.
4. Smaller footprint - Less clutter in the project, which is especially important for shared libraries.

#### Related Issues

fixes: #660

### DIFF
--- a/packages/espree/package.json
+++ b/packages/espree/package.json
@@ -46,7 +46,6 @@
     "@rollup/plugin-node-resolve": "^15.3.0",
     "eslint-release": "^3.2.0",
     "esprima-fb": "^8001.2001.0-dev-harmony-fb",
-    "npm-run-all2": "^6.2.2",
     "rollup": "^2.79.1",
     "shelljs": "^0.8.5"
   },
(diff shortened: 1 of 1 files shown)


nzakas (approved):
LGTM. Thanks.

PR #667, opened 2025-07-10
feat: add support for types for `espree`

lumirlumir:
#### Prerequisites checklist

-   [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What changes did you make? (Give an overview)

Working in progress.

#### Related Issues

Fixes: https://github.com/eslint/js/issues/529

PR #705, opened 2025-10-27
feat: TypeScript support for Espree

brettz9:
#### Prerequisites checklist

Provide TypeScript support for `espree`.

- applied `checkJs`/`allowJs` TypeScript to `espree`.
- complete test coverage
- in `lib/espree.js`, removes an uncovered and apparently unnecessary branch (previously lines 152-154) where a property was being added to an array.

#### Related Issues

A renewed approach to #544 which doesn't require a special, fragile build routine or non-standard JSDoc tags.

The `@typedef` exports that were of previous concern should not be here since the main file does not use them except for public type exports. Otherwise, we are using `@import`.

Has some added complexity in redefining Acorn/acorn-jsx because Acorn's TypeScript does not concern all properties of relevance to plugin authors. See https://github.com/acornjs/acorn/issues/1404 .

Currently applies acorn-jsx from my own fork as waiting on https://github.com/acornjs/acorn-jsx/pull/139 .

### DIFF
--- a/packages/espree/espree.js
+++ b/packages/espree/espree.js
@@ -69,7 +69,7 @@ import { getLatestEcmaVersion, getSupportedEcmaVersions } from "./lib/options.js
 // Types exported from file
 // ----------------------------------------------------------------------------
 /**
- * @typedef {acorn.ecmaVersion} EcmaVersion
+ * @typedef {3|5|6|7|8|9|10|11|12|13|14|15|16|17|2015|2016|2017|2018|2019|2020|2021|2022|2023|2024|2025|2026|'latest'} EcmaVersion
  */
 
 /**
--- a/packages/espree/CONTRIBUTING.md
+++ b/packages/espree/CONTRIBUTING.md
@@ -10,7 +10,7 @@ Our full contribution guidelines can be found at:
 # How to upgrade `acorn` to support new syntax
 
 1. `npm install acorn@latest`
-1. If a new `ecmaVersion` value is added, update `SUPPORTED_VERSIONS` constant in `lib/options.js` and tests in `tests/lib/supported-ecmaversions.js`.
+1. If a new `ecmaVersion` value is added, update the `SUPPORTED_VERSIONS` constant in `lib/options.js`, the tests in `tests/lib/supported-ecmaversions.js`, and the `EcmaVersion` type.
 1. If new token types are added, update `lib/token-translator.js` file to translate the tokens.
 1. Add tests in `tests/fixtures/ecma-version/<ecma-vesion>/`.
     - Add a directory named the new syntax name.


fasttime:
Not stale. This PR needs to be reviewed.

nzakas:
@brettz9 thanks for this. We're in the middle of preparing to release the first prerelease of v10, so we're a bit swamped at the moment. We'll get around to reviewing this once our schedules free up a bit.

packages/espree/package.json
```diff
@@ -37,15 +38,17 @@
   "license": "BSD-2-Clause",
   "dependencies": {
     "acorn": "^8.15.0",
-    "acorn-jsx": "^5.3.2",
+    "acorn-jsx": "brettz9/acorn-jsx#esm",
```
fasttime: I'm not sure if Acorn-JSX is still maintained. The last commit in that repo was three years ago. If the maintainer doesn't respond to acornjs/acorn-jsx#137, perhaps we should start considering a different approach for those types?

packages/espree/package.json
```diff
@@ -56,7 +59,9 @@
     "acorn"
   ],
   "scripts": {
-    "build": "rollup -c rollup.config.js",
+    "attw": "attw --pack",
```
fasttime: In the `eslint` repo, we use `lint:types` as a script name to run `attw`. I'd suggest doing the same here for consistency.

```suggestion
    "lint:types": "attw --pack",
```

packages/espree/espree.js
```diff
@@ -55,6 +55,59 @@
...
+ *     jsx?: boolean,
+ *     globalReturn?: boolean,
+ *     impliedStrict?: boolean
+ *   },
+ *   range?: boolean,
+ *   loc?: boolean,
+ *   tokens?: boolean,
+ *   comment?: boolean,
+ * }} ParserOptions
+ */
```
fasttime, line 133: `@types/espree` uses the name `Options` rather than `ParserOptions`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c807c15b6b56ef3f49f5543d228649e7dc64f14c/types/espree/index.d.cts#L40-L122

I'd suggest keeping `Options` for compatibility, to avoid introducing a breaking change.

packages/espree/espree.js
```diff
@@ -55,6 +55,59 @@
...
+/**
+ * @import { EnhancedSyntaxError } from "./lib/espree.js";
+ * @import { EspreeParserCtor, EspreeParserJsxCtor } from "./lib/types.js";
+ */
+
+// ----------------------------------------------------------------------------
+// Types exported from file
+// ----------------------------------------------------------------------------
+/**
+ * @typedef {3|5|6|7|8|9|10|11|12|13|14|15|16|17|2015|2016|2017|2018|2019|2020|2021|2022|2023|2024|2025|2026|'latest'} ecmaVersion
```
fasttime: In ESLint and in the core types, we use `EcmaVersion` for the type name, with the first letter capitalized:

https://github.com/eslint/rewrite/blob/core-v1.0.0/packages/core/src/types.ts#L832-L859

packages/espree/lib/token-translator.js
```diff
@@ -63,59 +108,70 @@ function convertTemplatePart(tokens, code) {
...
+/* eslint-disable jsdoc/check-types -- The API allows either */
 /**
  * Contains logic to translate Acorn tokens into Esprima tokens.
- * @param {Object} acornTokTypes The Acorn token types.
- * @param {string} code The source code Acorn is parsing. This is necessary
- *      to correct the "value" property of some tokens.
- * @constructor
  */
-function TokenTranslator(acornTokTypes, code) {
+class TokenTranslator {
```
fasttime: Makes sense to use a class here 👍

packages/espree/lib/options.js
```diff
@@ -3,11 +3,15 @@
  * @author Kai Cataldo
  */
 
+/**
+ * @import { ecmaVersion, ParserOptions } from "../espree.js";
```
fasttime: `ecmaVersion` looks unused.

```suggestion
 * @import { ParserOptions } from "../espree.js";
```

packages/espree/lib/token-translator.js
```diff
@@ -3,6 +3,50 @@
  * @author Nicholas C. Zakas
  */
 
+/**
+ * @import * as acorn from "acorn";
+ * @import { EnhancedTokTypes } from "./espree.js"
+ * @import { ecmaVersion } from "../espree.js";
```
fasttime: This is also unused.

```suggestion
```

packages/espree/lib/espree.js
```diff
@@ -12,15 +83,17 @@ const ESPRIMA_FINISH_NODE = Symbol("espree's esprimaFinishNode");
...
- * @param {int} end The index at which the comment ends.
- * @param {Location} startLoc The location at which the comment starts.
- * @param {Location} endLoc The location at which the comment ends.
- * @param {string} code The source code being parsed.
- * @returns {Object} The comment object.
+ * @param {Integer} start The index at which the comment starts.
+ * @param {Integer} end The index at which the comment ends.
+ * @param {acorn.Position | undefined} startLoc The location at which the comment starts.
+ * @param {acorn.Position | undefined} endLoc The location at which the comment ends.
+ * @param {string | string} code The source code being parsed.
```
fasttime: Maybe an accidental change?

```suggestion
 * @param {string} code The source code being parsed.
```

packages/espree/lib/token-translator.js
```diff
@@ -42,19 +86,20 @@ function convertTemplatePart(tokens, code) {
     const firstToken = tokens[0],
         lastTemplateToken = tokens.at(-1);
```
fasttime: Instead of changing the logic to handle cases where `lastTemplateToken` is undefined, which is unexpected, could we apply a type cast to exclude `undefined` from the inferred type?

In [`eslint/rewrite`](https://github.com/eslint/rewrite), we only use `strictNullChecks` to validate `.d.ts` files. For the JavaScript sources, that option is disabled because it's a pain to work with, and the lack of non-null assertions in JSDoc doesn't make things easier.

fasttime:
I think this is a very good approach to autogenerating the types. We could adopt some of these techniques in other repositories to simplify complexity.

brettz9:
> I think this is a very good approach to autogenerating the types. We could adopt some of these techniques in other repositories to simplify complexity.

Great... If you end up using this approach, I have a branch ready for eslint-scope as well (though it looks like you might already be working on one, sorry!).

fasttime:
> > I think this is a very good approach to autogenerating the types. We could adopt some of these techniques in other repositories to simplify complexity.
> 
> Great... If you end up using this approach, I have a branch ready for eslint-scope as well (though it looks like you might already be working on one, sorry!).

We can definitely look into autogenerating types for `eslint-scope` as well if we can maintain compatibility with the current types in `@types/eslint-scope`. I'd suggest doing that in a separate pull request once #709 is merged in order to avoid overlaps.

packages/espree/package.json
brettz9: Sure. I can look to add the types within this project by the time this may be deemed ready.

packages/espree/lib/token-translator.js
brettz9: Sure, I've cast it to `acorn.Token` without the `undefined`.

brettz9:
> Since `espree` types are a new feature, what do you think about adding tests for it?

Sure.

> We already have type tests for `eslint-visitor-keys` in [`packages/eslint-visitor-keys/test-d/index.test-d.ts`](https://github.com/eslint/js/blob/main/packages/eslint-visitor-keys/test-d/index.test-d.ts) and we could add similar tests for `espree` as well. DefinitelyTyped also has [type tests for `@types/espree`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/espree/espree-tests.ts) which could be a useful reference, if we want to build on their work.

That was a helpful pattern to follow. I've added a1d2f7a which adds their tests as well as 138f496e0c9431ca37ff55ed0c06817927db58b9 which adds the public exports they were missing. Let me know if you think we need coverage for other types (and which ones).

packages/espree/lib/token-translator.js
```diff
@@ -33,28 +76,29 @@ const Token = {
...
+        lastTemplateToken = /** @type {acorn.Token} */ (tokens.at(-1));
 
+    /** @type {EsprimaToken} */
     const token = {
         type: Token.Template,
         value: code.slice(firstToken.start, lastTemplateToken.end)
     };
 
-    if (firstToken.loc) {
+    if (firstToken.loc && lastTemplateToken.loc) {
```
fasttime: Instead of adding logic to check for `&& lastTemplateToken.loc`, can we use a type cast to access the property value below?

packages/espree/lib/token-translator.js
```diff
@@ -33,28 +76,29 @@ const Token = {
...
         token.loc = {
             start: firstToken.loc.start,
             end: lastTemplateToken.loc.end
         };
     }
 
-    if (firstToken.range) {
+    if (firstToken.range && lastTemplateToken.range) {
```
fasttime: As above, we could avoid checking for `&& lastTemplateToken.range` and just use a type cast to handle `lastTemplateToken.range` as non-undefined.

packages/espree/package.json
```diff
@@ -56,13 +60,16 @@
...
+    "tsc": "tsc && tsc -p tsconfig-cjs.json",
+    "build": "rollup -c rollup.config.js && npm run tsc",
     "build:debug": "npm run build -- -m",
     "build:docs": "node tools/sync-docs.js",
     "build:update-version": "node tools/update-version.js",
     "prepublishOnly": "npm run build:update-version && npm run build",
     "pretest": "npm run build",
-    "test": "npm run test:cjs && npm run test:esm",
+    "test": "npm run test:tsd && npm run test:cjs && npm run test:esm",
+    "test:tsd": "tsd --typings dist/espree.d.ts",
```
fasttime: I'd suggest calling the script `test:types`, as in `eslint-visitor-keys` and other ESLint packages.

```suggestion
    "test": "npm run test:types && npm run test:cjs && npm run test:esm",
    "test:types": "tsd --typings dist/espree.d.ts",
```

packages/espree/test-d/index.test-d.ts
```diff
@@ -0,0 +1,59 @@
...
+expectType<acorn.Node>(ast_option);
+
+const tokens = espree.tokenize("let foo = \"bar\"");
+expectType<EspreeTokens>(tokens);
+
+const tokens_option = espree.tokenize("let foo = \"bar\"", { ecmaVersion: 6 });
+expectType<EspreeTokens>(tokens_option);
+
+const name = espree.name;
+expectType<string>(name);
```
fasttime: I think this should be either:

```suggestion
name satisfies string;
```

or:


```suggestion
expectType<"espree">(name);
```

Same for the other failing assertions below.

packages/espree/espree.js
fasttime, line 1: When I run `npm run prepublishOnly`, I can see in `dist/espree.d.ts` that `version` has the type `"11.0.0"`.

Can we make sure that `version` has the type `string`, so that it stays unchanged upon releases?

https://github.com/eslint/js/blob/138f496e0c9431ca37ff55ed0c06817927db58b9/packages/espree/espree.js#L237

packages/espree/test-d/index.test-d.ts
fasttime: I think this should be either:

```suggestion
name satisfies string;
```

or:


```suggestion
expectType<"espree">(name);
```

And similarly for the other failing assertions below.

packages/espree/tsconfig.json
fasttime, line 1: I'm not sure if it's necessary to generate type definitions for all `lib` files. From what I can see, `dist/espree.d.ts` only imports `EsprimaToken` from `token-translator.js` and `EspreeComment` from `espree.js`. If those two types could be inlined somehow, maybe we could avoid generating and distributing type definitions for the entire `lib` directory.

fasttime:
> > We already have type tests for `eslint-visitor-keys` in [`packages/eslint-visitor-keys/test-d/index.test-d.ts`](https://github.com/eslint/js/blob/main/packages/eslint-visitor-keys/test-d/index.test-d.ts) and we could add similar tests for `espree` as well. DefinitelyTyped also has [type tests for `@types/espree`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/espree/espree-tests.ts) which could be a useful reference, if we want to build on their work.
> 
> That was a helpful pattern to follow. I've added [a1d2f7a](https://github.com/eslint/js/commit/a1d2f7a90e4221544da53e5ea5273c5105c240cd) which adds their tests as well as [138f496](https://github.com/eslint/js/commit/138f496e0c9431ca37ff55ed0c06817927db58b9) which adds the public exports they were missing. Let me know if you think we need coverage for other types (and which ones).

I think that's good enough for a start.

packages/espree/tsconfig.json
brettz9: I was able to move EspreeComment/EspreeToken into the main export file. And for ESM types, I was able to use `outFile` in the `tsconfig.json` to ensure that only a single file was built without `lib` being added.

However, there appear to be at least two problems here:

1. The generated types look poorly formed to me. For example, there is an import of "espree.js" without a slash at the beginning. I don't know if this will work in production even though it was produced by `tsc` without errors.
2. @arethetypeswrong/cli reports an error now with the CJS types. Using `outFile`, I am not able to change the file extension to end in `d.cts`, and targeting the `d.ts` file it generates leads to errors by the lint:types script. With `outDir`, the `lib` directory is generated.

I've added the commit so you can see, but I think we should probably revert it unless you know of a way to workaround.

packages/espree/espree.js
```diff
@@ -55,6 +55,72 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+/**
+ * @import {
+ *   EspreeParserCtor, EspreeParserJsxCtor, BaseEsprimaToken
```
fasttime: Can we inline `BaseEsprimaToken` as well, since it's only used in the type definition of `EspreeToken` in this file?

packages/espree/tsconfig.json
fasttime: If it's not possible to generate only types for `espree.js` without also including `lib` or bundling the output with `outFile`, then maybe another option is using `outDir` like it was done previously, and delete `dist/lib` afterwards to avoid including the `.d.ts` files for `lib` in the package distribution. This will also ensure that those files aren't required to test the types. What do you think?

brettz9: Regarding your comments in https://github.com/eslint/js/pull/705#discussion_r2584552124 , being as there are no longer any type dependencies relying on `lib` in the main file, I think that is viable, and I've added such a commit to remove that `dist/lib` directory at the end of `build`.

> This will also ensure that those files aren't required to test the types.

Sure. While the imports weren't reaching into the `dist` folder before my commits just now (as the imported ESM file was in source being as there is no generated JS-from-TS ESM file to target), I did take the spirit of your suggestion and change to import the CJS file in `dist` (being in Node, we can import a CJS file in our test file). That helped reveal that there was one type issue also fixed now in the same commit (that `VisitorKeys` had to have an explicit type annotation to avoid import path issues). Now there are no errors when viewing the `dist` declaration files.

packages/espree/package.json
```diff
@@ -56,13 +61,16 @@
     "acorn"
   ],
   "scripts": {
-    "build": "rollup -c rollup.config.js",
+    "lint:types": "attw --pack",
+    "tsc": "tsc && tsc -p tsconfig-cjs.json",
+    "build": "rollup -c rollup.config.js && npm run tsc && rm -Rf dist/lib",
```
fasttime: It's better to use JavaScript code to delete the directory because `rm -Rf` doesn't work in PowerShell. You could for example create a script in `packages/espree/tools`. An example is in the `sync-docs` tool:

https://github.com/eslint/js/blob/6f4d32568c247978153105af2f2a59989abc55d7/packages/espree/tools/sync-docs.js#L26

Another option is using `node` with inline code, for example `node -e "fs.rmSync('dist/lib', { recursive: true })"`. We are doing something similar in the rewrite package to copy a directory:

https://github.com/eslint/rewrite/blob/0f16298e06e0380f259611c7786a195a4627ee34/packages/core/package.json#L20

packages/espree/tsconfig.json
fasttime: Thanks! I was actually thinking one step ahead: in some other packages we use type tests to ensure that the CommonJS types can be imported, in addition to the regular type tests. For example, see `@eslint/json`:

https://github.com/eslint/json/blob/main/tests/types/cjs-import.test.cts

My thought was that we could do the same in Espree to verify the CJS types. However, since we're now using `attw`, it might be redundant to check that the type definitions can be imported in different formats, so your approach seems sufficient.

packages/espree/package.json
fasttime: If I revert this change on my machine, the build succeeds and all tests pass. Do you think we could drop the `acorn-jsx` update from this PR?

brettz9: If you remove the `acorn-jsx` folder in `node_modules` and then run `npm i`, it won't work. However, I've added a commit now to inline those desired acorn-jsx-based types, so the PR no longer depends on my fork.

fasttime: I see now that the `tsc` build for `eslint-visitor-keys` fails if the `acorn-jsx` types aren't patched. I also think inlining those types makes the most sense.

packages/espree/package.json
```diff
@@ -56,13 +61,16 @@
     "acorn"
   ],
   "scripts": {
-    "build": "rollup -c rollup.config.js",
+    "lint:types": "attw --pack",
```
fasttime, line 61: Just a note that #709 enables `attw` in CI for all packages:

https://github.com/eslint/js/blob/3658660c0bbc4878498ae3f25c3c8f6cc3f029e4/.github/workflows/ci.yml#L59-L60
https://github.com/eslint/js/blob/3658660c0bbc4878498ae3f25c3c8f6cc3f029e4/package.json#L11

This means that if that change is merged before this PR, we'll need to ensure the Espree types are generated prior to running `attw`. This could be done either by extending the npm script itself (`lint:types`), or in the CI job.

fasttime:
Note: I've renamed the PR to better reflect the intent in the changelog.

packages/espree/lib/espree.js
```diff
@@ -12,15 +67,17 @@ const ESPRIMA_FINISH_NODE = Symbol("espree's esprimaFinishNode");
...
```
lumirlumir:
```suggestion
 * @param {string} code The source code being parsed.
```

Question: Is this a typo, or was this type description intentional?

packages/espree/lib/espree.js
```diff
@@ -12,15 +67,17 @@ const ESPRIMA_FINISH_NODE = Symbol("espree's esprimaFinishNode");
...
  * @param {boolean} block True if it's a block comment, false if not.
  * @param {string} text The text of the comment.
- * @param {int} start The index at which the comment starts.
- * @param {int} end The index at which the comment ends.
- * @param {Location} startLoc The location at which the comment starts.
- * @param {Location} endLoc The location at which the comment ends.
- * @param {string} code The source code being parsed.
- * @returns {Object} The comment object.
+ * @param {Integer} start The index at which the comment starts.
+ * @param {Integer} end The index at which the comment ends.
```
lumirlumir:
```suggestion
 * @param {number} start The index at which the comment starts.
 * @param {number} end The index at which the comment ends.
```

Could we use `number` here instead of `Integer`? The other types consistently use `number`, so it seems to be more consistent.

brettz9:
> Hi, I have a question about this PR!
> 
> Could this PR fully replace `@types/espree`?

Yes, that is the plan, and the approach was confirmed in https://github.com/eslint/js/issues/529#issuecomment-1013389654 .

> I was planning to work on #667 (though it hasn't progressed yet 😅). If this PR contains all the context mentioned in #529, I plan to close #667.

If by all the context, you mean the idea of staying with Rollup, yes, this stays with Rollup and provides separate types for ESM/CJS at `./dist/espree.d.ts` and `/dist/espree.d.cts`.

packages/espree/lib/espree.js
brettz9: Though it makes no difference to TypeScript, it does convey the intention more clearly to developers and users.

IIRC, I believe @nzakas favored this approach as well.

If it's desired, I can change the other relevant instances of `number` to `Integer` as well, but didn't want to add too much diff noise at once.

nzakas: Let's use `number`.

brettz9:
> I’m sorry for the delay. 

It's all good!

> Could you take a look at the merge conflict when you have a moment?

Sure! Rebased...

packages/espree/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "lib": ["ESNext"],
+    "moduleResolution": "nodenext",
+    "module": "nodenext",
+    "allowJs": true,
+    "checkJs": true,
+    "noEmit": false,
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "strict": true,
```
lumirlumir, line 12:
```suggestion
    "strict": true,
    "exactOptionalPropertyTypes": true,
```

Could we also add `"exactOptionalPropertyTypes": true`?

The `strict` mode doesn't include `exactOptionalPropertyTypes` (ref: https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes), and there have been type issues when not enabling it in other packages — for example, https://github.com/eslint/markdown/issues/402. 

Also, the `rewrite`, `css`, `json`, and `markdown` repositories have enabled `exactOptionalPropertyTypes` for their type tests.

- https://github.com/eslint/css/blob/main/tests/types/tsconfig.json#L7
- https://github.com/eslint/json/blob/main/tests/types/tsconfig.json#L7
- https://github.com/eslint/markdown/blob/main/tests/types/tsconfig.json#L8
- https://github.com/eslint/rewrite/blob/main/packages/core/tests/types/tsconfig.json#L4

So, I think it would be helpful if we can add it.

packages/espree/tsconfig-cjs.json
```diff
@@ -0,0 +1,18 @@
...
+    "lib": ["ESNext"],
+    "moduleResolution": "node",
+    "module": "commonjs",
+    "allowJs": true,
+    "esModuleInterop": true,
+    "noEmit": false,
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "strict": true,
```
lumirlumir, line 12:
```suggestion
    "strict": true,
    "exactOptionalPropertyTypes": true,
```

Same as the comment above!

packages/espree/test-d/index.test-d.ts
```diff
@@ -0,0 +1,60 @@
+import { expectType, expectAssignable } from 'tsd';
+
+import * as espree from "../dist/espree.cjs";
+import * as acorn from "acorn";
+import type { Options, EspreeTokens } from "../dist/espree.cjs";
+import type {VisitorKeys} from "../../eslint-visitor-keys/lib/index.js";
```
lumirlumir:
```suggestion
import type { VisitorKeys } from "eslint-visitor-keys";
```

Just a small note: since `eslint-visitor-keys` is also located under the `packages` directory and is handled by npm workspaces (as shown below), it will be symbolically linked in `node_modules`, so I think we can simply use `eslint-visitor-keys` instead of `../../eslint-visitor-keys/lib/index.js`.

https://github.com/eslint/js/blob/5598f8895a3f21fc140a0be97fdf207de425854a/package.json#L12-L14

packages/espree/lib/options.js
```diff
@@ -22,19 +26,23 @@ const SUPPORTED_VERSIONS = [
...
+ */
 
 /**
  * Get the latest ECMAScript version supported by Espree.
- * @returns {number} The latest ECMAScript version.
+ * @returns {normalizedEcmaVersion} The latest ECMAScript version.
  */
 export function getLatestEcmaVersion() {
-    return SUPPORTED_VERSIONS.at(-1);
+    return /** @type {normalizedEcmaVersion} */ (SUPPORTED_VERSIONS.at(-1));
```
lumirlumir:
```diff
+const LATEST_ECMA_VERSION =
+   /** @type {typeof SUPPORTED_VERSIONS extends readonly [...infer _R, infer L] ? L : never} */ (
+       SUPPORTED_VERSIONS.at(-1)
+   );

/**
 * Get the latest ECMAScript version supported by Espree.
+* @returns {typeof LATEST_ECMA_VERSION} The latest ECMAScript version.
 */
export function getLatestEcmaVersion() {
+   return LATEST_ECMA_VERSION;
}
```

image

Since `getLatestEcmaVersion` always returns the last item in `SUPPORTED_VERSIONS` (now `17`), it would be nice if we could infer its return type as the exact value `17` rather than a union like `3 | 5 | 6 | 7 | 8 | 9 | 10 | ...`.

https://github.com/eslint/js/blob/5598f8895a3f21fc140a0be97fdf207de425854a/packages/espree/lib/options.js#L31-L33

I tried creating a helper type; an approach like `typeof SUPPORTED_VERSIONS extends readonly [...infer _R, infer L] ? L : never` allows the type to be inferred automatically.

I think it would be helpful to make the type more specific, either by using the approach above or by another approach if you prefer.

packages/espree/espree.js
```diff
@@ -123,9 +236,9 @@ export function tokenize(code, options) {
 /**
  * Parses the given code.
  * @param {string} code The code to tokenize.
- * @param {Object} options Options defining how to tokenize.
- * @returns {ASTNode} The "Program" AST node.
- * @throws {SyntaxError} If the input code is invalid.
+ * @param {Options} [options] Options defining how to tokenize.
+ * @returns {acorn.Node} The "Program" AST node.
```
lumirlumir:
```suggestion
 * @returns {acorn.Program} The "Program" AST node.
```

Image

Question:

Would there be a reason why we return `acorn.Node` instead of `acorn.Program`?

It seems the JSDoc specifies the return type as "Program AST node", as shown below:

https://github.com/eslint/js/blob/5598f8895a3f21fc140a0be97fdf207de425854a/packages/espree/espree.js#L127

Also, the `@types/espree` package specifies that the `parse` function returns `acorn.Program`:

https://www.npmjs.com/package/@types/espree?activeTab=code

Image

Since `acorn.Program` derives from `acorn.Node` and is a bit more specific, I am wondering:

Image

packages/espree/test-d/index.test-d.ts
```diff
@@ -0,0 +1,60 @@
+import { expectType, expectAssignable } from 'tsd';
+
+import * as espree from "../dist/espree.cjs";
+import * as acorn from "acorn";
+import type { Options, EspreeTokens } from "../dist/espree.cjs";
```
lumirlumir:
```suggestion
import * as espree from "../dist/espree.cjs";
import * as acorn from "acorn";
```

Maybe, we could also simplify the import by using `espree.Options` and `espree.EspreeTokens`, similar to using `acorn.Node`, instead of destructuring them from the same import source (`../dist/espree.cjs`). What do you think?

lumirlumir (changes requested):

packages/espree/lib/options.js
brettz9: Very nice, thanks!

packages/espree/espree.js
brettz9: Great, thanks! Committed...

packages/espree/test-d/index.test-d.ts
brettz9: Works for me... Applied...

packages/espree/lib/espree.js
```diff
@@ -3,6 +3,60 @@
...
+/**
+ * @import {
+ *   CommentType,
+ *   EspreeParserCtor,
+ *   EsprimaNode,
+ *   AcornJsxParserCtorEnhanced,
+ *   TokTypes
+ * } from "./types.js";
+ * @import { EspreeToken as EsprimaToken } from "../espree.js";
+ * @import { Options, EspreeComment as EsprimaComment } from "../espree.js";
```
lumirlumir:
```suggestion
 * @import {
 *   Options,
 *   EspreeToken as EsprimaToken,
 *   EspreeComment as EsprimaComment
 * } from "../espree.js";
```

Perhaps, we could consolidate the imports into a single statement?

packages/espree/espree.js
```diff
@@ -55,6 +55,76 @@
...
+/**
+ * @import { EspreeParserCtor, EspreeParserJsxCtor } from "./lib/types.js";
+ */
+
+// ----------------------------------------------------------------------------
+// Types exported from file
+// ----------------------------------------------------------------------------
+/**
+ * @typedef {3|5|6|7|8|9|10|11|12|13|14|15|16|17|2015|2016|2017|2018|2019|2020|2021|2022|2023|2024|2025|2026|'latest'} EcmaVersion
+ */
```
lumirlumir, line 73: ~Could we also update the section below in `CONTRIBUTING.md` to mention updating the `EcmaVersion` type when a new `ecmaVersion` value is added?~

https://github.com/eslint/js/blob/5598f8895a3f21fc140a0be97fdf207de425854a/packages/espree/CONTRIBUTING.md?plain=1#L13

(Please ignore this comment if the approach below is more acceptable.)

packages/espree/lib/options.js
```diff
@@ -22,19 +26,30 @@ const SUPPORTED_VERSIONS = [
...
  */
 export function getLatestEcmaVersion() {
-    return SUPPORTED_VERSIONS.at(-1);
+    return LATEST_ECMA_VERSION;
 }
 
 /**
  * Get the list of ECMAScript versions supported by Espree.
- * @returns {number[]} An array containing the supported ECMAScript versions.
+ * @returns {normalizedEcmaVersion[]} An array containing the supported ECMAScript versions.
```
lumirlumir:
```suggestion
 * @returns {[...typeof SUPPORTED_VERSIONS]} An array containing the supported ECMAScript versions.
```

Like in the comment https://github.com/eslint/js/pull/705#discussion_r2643219478, I think it would be nice to narrow the return type of `getSupportedEcmaVersions` slightly.

Currently, the return type of `getSupportedEcmaVersions()` is typed as follows:

Image

```ts
(3 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17)[]
```

This means an array like `[3, 3, 3, 3, 3, 3, 3]` can satisfy `(3 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17)[]`.

However, if we use `[...typeof SUPPORTED_VERSIONS]` as the return type, the type can be expressed more precisely, like below:

Image

```ts
[3, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17]
```

Since the `getSupportedEcmaVersions()` function returns the `SUPPORTED_VERSIONS` array items unchanged, I think that would work. What do you think?

packages/espree/espree.js
lumirlumir:
```suggestion
/**
 * @typedef {acorn.ecmaVersion} EcmaVersion
 */
```

After further investigation, I think another simple approach—besides the one mentioned above—would be to use `acorn.ecmaVersion` instead of updating the type manually.

`acorn.ecmaVersion` is updated automatically as `acorn` gains support for newer ECMAScript syntax, so manual updates would no longer be necessary.

Image

packages/espree/lib/types.js
```diff
@@ -0,0 +1,118 @@
...
+ * } & acorn.Node} EsprimaNode
+ */
+
+/**
+ * @typedef {"Block"|"Hashbang"|"Line"} CommentType
+ */
+
+/**
+ * @typedef {{
+ *   tokenize: () => import('../espree.js').EspreeToken[],
```
lumirlumir:
```suggestion
 *   tokenize: () => EspreeToken[],
```

Image

Perhaps we could remove the inline `import('../espree.js')` and move it to a top-level `@import` for consistency?

packages/espree/lib/espree.js
```diff
@@ -3,6 +3,60 @@
...
+ *   sourceType?: "script"|"module"|"commonjs";
+ *   comments?: EsprimaComment[];
+ *   tokens?: EsprimaToken[];
+ *   body: acorn.Node[];
+ * } & acorn.Program} EsprimaProgramNode
+ */
+
+// ----------------------------------------------------------------------------
+// Types exported from file
+// ----------------------------------------------------------------------------
```
lumirlumir, line 47:
```suggestion
```

Question:

I'm not 100% sure, but are `EnhancedSyntaxError` and `EnhancedTokTypes` still exported?

They're mentioned in the `@throws` JSDoc for the `tokenize` and `parse` functions in `packages/espree/espree.js`, but that doesn't seem to affect using them as types.

It seems `@throws {EnhancedSyntaxError}` is purely documentation and has no type-related effect:

Image

lumirlumir (changes requested):
Thanks for your hard work on the reviews! 

I've left a few more comments throughout.

packages/espree/lib/types.js
brettz9: Sure--I've added a commit to do so...

packages/espree/lib/espree.js
brettz9: `EnhancedSyntaxError` is not currently imported from outside of this file, though it probably should be imported in `/espree.js` for use within its `@throws` so that VSCode will show the appropriate type (though there is currently no tsc error there).

`EnhancedTokTypes` is imported by `token.translator.js` for use in an actual type (within the constructor for the `TokenTranslator` class).

In my view, it is still accurate to suggest they are exported from the file, even if they're not exported from the package.

Let me know though if you'd like me to either remove the comment block or add the `EnhancedSyntaxError` import.

lumirlumir: Thanks for the explanation! If that's the case, I think it's fine to leave it as is.

packages/espree/espree.js
```diff
@@ -137,10 +250,14 @@ export function parse(code, options) {
 // Public
 //------------------------------------------------------------------------------
 
+/** @type {string} */
 export const version = "11.0.0"; // x-release-please-version
 export const name = "espree";
 
 /* istanbul ignore next */
+/**
+ * @type {import("eslint-visitor-keys").VisitorKeys}
```
lumirlumir:
```suggestion
 * @type {visitorKeys.VisitorKeys}
```

Non-blocking: 

Maybe we could simplify this using `visitorKeys`, since we're already importing `import * as visitorKeys from "eslint-visitor-keys";` on line 132 :)

lumirlumir (approved):
LGTM, thanks! I've left a single optional comment.

Would like @fasttime to verify the changes before merging.

lumirlumir:
FYI: 

I'm closing https://github.com/eslint/js/pull/667 in favor of this PR and linking issue https://github.com/eslint/js/issues/529, since this PR resolves the issue originally intended to be fixed by #667.

packages/espree/lib/options.js
```diff
@@ -22,19 +26,30 @@ const SUPPORTED_VERSIONS = [
...
-];
+]);
+
+/**
+ * @typedef {typeof SUPPORTED_VERSIONS[number]} normalizedEcmaVersion
+ */
+
+const LATEST_ECMA_VERSION =
+    /* eslint-disable jsdoc/valid-types -- Bug */
+    /** @type {typeof SUPPORTED_VERSIONS extends readonly [...infer _R, infer L] ? L : never} */ (
```
fasttime:
```suggestion
    /** @type {typeof SUPPORTED_VERSIONS extends readonly [...unknown[], infer L] ? L : never} */ (
```

since `_R` is unused.

packages/espree/espree.js
```diff
@@ -55,6 +55,76 @@
...
+ *     jsx?: boolean,
+ *     globalReturn?: boolean,
+ *     impliedStrict?: boolean
+ *   },
+ *   range?: boolean,
+ *   loc?: boolean,
+ *   tokens?: boolean,
+ *   comment?: boolean,
+ * }} Options
+ */
```
fasttime: I think our convention is to add JSDoc types after imports:

https://github.com/eslint/js/blob/b02cfa99d8ceab25aed9afb14f392e83ac9a31c4/packages/espree/espree.js#L129-L133

packages/espree/espree.js
```diff
@@ -55,6 +55,76 @@
...
+/**
+ * @import { EspreeParserCtor, EspreeParserJsxCtor } from "./lib/types.js";
+ */
+
+// ----------------------------------------------------------------------------
+// Types exported from file
+// ----------------------------------------------------------------------------
+/**
+ * @typedef {acorn.ecmaVersion} EcmaVersion
+ */
```
fasttime, line 73: I'm not sure if it's very good to define `EcmaVersion` the same way as in Acorn. Acorn adds support for new ECMA versions in minor releases (for example, ES 2026 was added in Acorn v8.15.0), meaning that ESLint gets updated to these new releases automatically. But Espree needs work before a new ECMA version can be supported. See for example #658.

@lumirlumir what do you think?

packages/espree/lib/options.js
```diff
@@ -22,19 +26,30 @@ const SUPPORTED_VERSIONS = [
     15, // 2024
     16, // 2025
     17 // 2026
-];
+]);
+
+/**
+ * @typedef {typeof SUPPORTED_VERSIONS[number]} normalizedEcmaVersion
+ */
```
fasttime, line 33: Can we name this `NormalizedEcmaVersion` (with a capital N) to follow our naming conventions?

packages/espree/lib/options.js
fasttime:
```suggestion
    /** @type {typeof SUPPORTED_VERSIONS extends readonly [...unknown[], infer L] ? L : never} */ (
```

since `_R` would be unused.

packages/espree/espree.js
fasttime, line 133: Oh, and could we move the JSDoc types after the static imports to reflect our conventions?

https://github.com/eslint/js/blob/b02cfa99d8ceab25aed9afb14f392e83ac9a31c4/packages/espree/espree.js#L129-L133

See also https://github.com/eslint/eslint/blob/v10.0.0-beta.0/.github/copilot-instructions.md#source-file-conventions.

packages/espree/espree.js
lumirlumir: > I'm not sure if it's very good to define EcmaVersion the same way as in Acorn. Acorn adds support for new ECMA versions in minor releases (for example, ES 2026 was added in Acorn v8.15.0), meaning that ESLint gets updated to these new releases automatically. But Espree needs work before a new ECMA version can be supported.

That makes sense. I have missed that point.

If so, would reverting that change and updating the section below in `CONTRIBUTING.md` to mention updating the `EcmaVersion` type when a new `ecmaVersion` value is added be a good idea?

https://github.com/eslint/js/blob/5598f8895a3f21fc140a0be97fdf207de425854a/packages/espree/CONTRIBUTING.md?plain=1#L13

lumirlumir:
Hi, @brettz9 

If you have a moment, could you take a look at the merge conflict and the comment in https://github.com/eslint/js/pull/705#discussion_r2649652159? I think that's the last comment left to resolve in this PR. 😄

packages/espree/espree.js
brettz9: @lumirlumir : That sounds fine to me. I didn't know if you wanted to await @fasttime 's reply, but I have attempted to implement your suggestion for `CONTRIBUTING.md` in addition to reverting `EcmaVersion` to be explicit about accepted number values.

PR #714, opened 2025-12-18
refactor: use release-please to update exported `version`

mdjermanovic:
#### Prerequisites checklist

Simplifies the way `eslint-scope` and `espree` are exporting `version`.

Removed prepublish step that updates `version`, and configured release-please to update it similar to how we're doing that in other projects (eslint/rewrite, eslint/json, eslint/markdown, eslint/css). Also added tests that should catch any problems with this when we manage to run CI on release-please PRs (https://github.com/eslint/js/pull/713).

### DIFF
--- a/packages/espree/tests/lib/version.test.js
+++ b/packages/espree/tests/lib/version.test.js
@@ -8,15 +8,10 @@
 //------------------------------------------------------------------------------
 
 import * as espree from "../../espree.js";
-import { fileURLToPath } from "node:url";
-import path from "node:path";
-import fs from "node:fs";
 import assert from "node:assert";
+import packageJson from "../../package.json" with { type: "json" };
 
-// eslint-disable-next-line no-underscore-dangle -- Conventional
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-
-const { version } = JSON.parse(fs.readFileSync(`${__dirname}/../../package.json`, "utf8"));
+const { version } = packageJson;
 
 //------------------------------------------------------------------------------
 // Tests


packages/eslint-scope/package.json
```diff
@@ -30,8 +30,6 @@
   "license": "BSD-2-Clause",
   "scripts": {
     "build": "rollup -c",
-    "build:update-version": "node tools/update-version.js",
-    "prepublishOnly": "npm run build:update-version && npm run build",
```
mdjermanovic, line 34: We're already running `build` before publishing:

https://github.com/eslint/js/blob/04247cbf4b4b99a8eb5389952df0af504814582d/.github/workflows/release-please.yml#L44-L47

packages/eslint-scope/tests/version.test.js
```diff
@@ -0,0 +1,29 @@
...
+import { expect } from "chai";
+import * as eslintScope from "../lib/index.js";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+import fs from "node:fs";
+
+// eslint-disable-next-line no-underscore-dangle -- Conventional
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const { version } = JSON.parse(fs.readFileSync(`${__dirname}/../package.json`, "utf8"));
```
fasttime: A note that this could be simplified:

```suggestion
import packageJson from "../package.json" with { type: "json" };

const { version } = packageJson;
```

Or, similarly:

```suggestion

const { default: { version } } = await import("../package.json", { with: { type: "json" } });
```

Per https://nodejs.org/api/esm.html#import-attributes, import attributes have been supported since Node.js v20.10.0.

packages/espree/tests/lib/version.test.js
```diff
@@ -0,0 +1,29 @@
...
+import * as espree from "../../espree.js";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+import fs from "node:fs";
+import assert from "node:assert";
+
+// eslint-disable-next-line no-underscore-dangle -- Conventional
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const { version } = JSON.parse(fs.readFileSync(`${__dirname}/../../package.json`, "utf8"));
```
fasttime: Same note as https://github.com/eslint/js/pull/714#discussion_r2634182521.

lumirlumir (approved):
LGTM, thanks! I don't have any further comments.

(I didn't move it to second review needed since there are pending comments.)

packages/espree/tests/lib/version.test.js
mdjermanovic: Updated in https://github.com/eslint/js/pull/714/commits/b08c10c5a68af2661396637058900a79a6d478a8

PR #740, opened 2026-03-26
chore: update TypeScript to v6

sethamus:
#### Prerequisites checklist

#### AI acknowledgment

- [x] I did _not_ use AI to generate this PR.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

This PR updates the repo’s TypeScript tooling to v6.

- bump `typescript` to v6 in `package.json`.
- updated the TypeScript module settings to use `moduleResolution: "nodenext"` and `module: "nodenext"` instead of the deprecated `moduleResolution: "node"` setup.
- added `rootDir: "lib"` to `eslint-visitor-keys`.


nzakas:
Please hold off. There's already a bunch of "Update TypeScript to v6" PRs across the org and we're not even sure we want to upgrade immediately.

sethamus:
@nzakas Sorry for the PRs! I opened these based on the discussion in my other PR (https://github.com/eslint/eslint/pull/20547#issuecomment-3941661887), which suggested that once TypeScript 6 is officially released, we should update our tooling.

nzakas:
Gotcha. @fasttime I'll leave it up to you to review.

package.json
lumirlumir, line 1: I think it would also be nice to update `@typescript-eslint/parser` in `eslint-scope` to the latest version, since it has supported TypeScript v6 since 8.58.0.

https://github.com/typescript-eslint/typescript-eslint/releases/tag/v8.58.0

https://github.com/eslint/js/blob/cf1e1d2efd2b22d15396a65b58e3a54079b5481a/packages/eslint-scope/package.json#L58

lumirlumir (changes requested):
Just a note that there are currently some CI failures related to this update.

fasttime:
The type tests are still failing.

tsconfig.base.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "allowImportingTsExtensions": true,
+    "allowJs": true,
+    "checkJs": true,
+    "declaration": true,
```
fasttime, line 5: Maybe we could set `declarationMap` here to keep the `.map` files that are currently generated in `eslint-visitor-keys/dist` and `espree/dist`. I'm not sure if we need declaration maps at all, but since those files are currently included in the bundle, some tools may rely on them.

```suggestion
    "declaration": true,
    "declarationMap": true,
```

packages/eslint-scope/tests/types/tsconfig.json
fasttime, line 1: The `strict` option also enables `noImplicitAny`, but with this change, `noImplicitAny` would be turned off in type tests. To restore the check we could add `"noImplicitAny": true` to this file, or we could change the parent config to `"../../../../tsconfig.base.json"`.

sethamus: Good catch. I've added `"noImplicitAny": true` here as suggested.

packages/eslint-scope/tests/types/cjs-import.test.cts
lumirlumir, line 1: Is there a reason why the named export test was removed?

This test verifies that the named export types exist correctly in the CJS export, so I think it need to be retained.

packages/eslint-visitor-keys/test-d/index.test-d.ts
```diff
@@ -1,6 +1,11 @@
 import { expectType, expectAssignable } from "tsd";
 
-import { KEYS, getKeys, unionWith, VisitorKeys } from "../";
+import {
+	KEYS,
+	getKeys,
+	unionWith,
+	VisitorKeys,
+} from "../dist/eslint-visitor-keys.cjs";
```
lumirlumir:
```suggestion
} from "eslint-visitor-keys";
```

I think simply using `eslint-visitor-keys` would be more accurate here, since the previous test using `../` resolved the type declaration file from the Module declaration surface rather than the CJS format. It was originally linked to `lib/index.js` and directed to `dist/index.d.ts`.

Additionally, using `eslint-visitor-keys` directly would be more beneficial, as it follows the same package-name type resolution path used by real consumers, since it is symlinked in `node_modules` and consumed like a real package.

lumirlumir (changes requested):

packages/eslint-scope/tests/types/cjs-import.test.cts
sethamus: We would otherwise get:
```
ECMAScript imports and exports cannot be written in a CommonJS file under 'verbatimModuleSyntax'.
```

fasttime:
[Are the types wrong?](https://github.com/eslint/js/actions/runs/25960118925/job/76343013673?pr=740) CI job is now failing with `attw --pack` crashing on `Cannot read properties of undefined (reading 'filename')`. This seems related to arethetypeswrong/arethetypeswrong.github.io#258, and may require the same temporary workaround used in eslint/eslint#20877 or eslint/rewrite#458.

fasttime:
To clear the CI failure: eslint/js#751.

packages/eslint-scope/tests/types/cjs-import.test.cts
fasttime: If we don't want to remove the CommonJS type tests, perhaps we could switch the `import` into an `import type` statement, to make sure at least that the type definitions are exported.

Another slightly more involved option is to validate `cts` files  without `verbatimModuleSyntax`.

sethamus: Yeah, we could switch it to an `import type` statement 👍.

fasttime (dismissed):
LGTM, thanks! Leaving open for @lumirlumir to review the changes.

lumirlumir:
I’m sorry for the delay. I need some dedicated time to review this PR, and I’ll make sure to review it by the end of this weekend.

tsconfig.base.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "allowImportingTsExtensions": true,
```
lumirlumir:
```suggestion

```

Is this option really necessary? It seems that this option wasn’t in any `tsconfig.json` before and was added here newly. This would allow looser type checking, as it allows `.ts` import syntax, so I think removing it here would be preferred.

tsconfig.base.json
```diff
@@ -0,0 +1,14 @@
...
+    "allowJs": true,
+    "checkJs": true,
+    "declaration": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": true,
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
lumirlumir:
```suggestion
    "module": "nodenext",
    "moduleResolution": "nodenext",
```

Non-blocking, but to match TypeScript documentation’s suggestion and the previous option value: https://www.typescriptlang.org/tsconfig/#module

tsconfig.base.json
```diff
@@ -0,0 +1,14 @@
...
+    "strict": true,
```
lumirlumir, line 9:
```suggestion
    "strict": true,
    "exactOptionalPropertyTypes": true,
```

`"exactOptionalPropertyTypes": true` is missing, while the previous `packages/eslint-scope/tsconfig.json`, `packages/eslint-visitor-keys/tsconfig.json`, and `packages/espree/tsconfig.json` included it.

packages/eslint-visitor-keys/tsconfig.json
lumirlumir, line 1: Though the behavior would be the same, I think updating `rollup-plugin-dts` to the latest version here as well like `@typescript-eslint/parser` would be helpful, since TS 6 support was added in `6.4.0`.

https://github.com/eslint/js/blob/59cba5c9ab49283af1a1609ec6ca6d56c7e777a1/packages/eslint-visitor-keys/package.json#L29

https://github.com/Swatinem/rollup-plugin-dts/blob/master/CHANGELOG.md#640

packages/espree/rollup.config.js
```diff
@@ -1,9 +1,22 @@
...
+		external: ["acorn", "acorn-jsx", "eslint-visitor-keys"],
+		treeshake: false,
+		output: {
+			file: "dist/espree.cjs",
+			format: "cjs",
+		},
 	},
-};
+	{
+		plugins: [dts()],
```
lumirlumir, line 14: After adding the `dts` plugin, I found that the `espree.d.cts.map` file is now missing.

Is there a way we can retain the previous behavior?

- Before:

https://npmx.dev/package-code/espree/v/11.2.0/dist

Image

- After:

Image

---

Or would removing the entire `.map` files under the `dist` directory be a better fit here?

After this TypeScript v6 change, other packages such as `eslint-scope` now emit `dist/eslint-scope.cjs.map`, and `eslint-visitor-keys` now emit `dist/index.d.ts.map` and `dist/visitor-keys.d.ts.map` files.

Especially the `dist/eslint-scope.cjs.map` file's size is quite large, so I’m not sure adding this map file here would be a good addition, since we usually don’t emit map files in other packages.

lumirlumir (changes requested):

packages/espree/rollup.config.js
sethamus: We’ve always generated map files for all packages, but we only publish them for `espree`. So for `eslint-scope` and `eslint-visitor-keys`, I think it’s fine to omit the `.map` files.

package.json
```diff
@@ -5,13 +5,19 @@
   "scripts": {
     "build": "npm run build --workspaces --if-present",
     "build:types": "npm run build:types --workspaces --if-present",
+    "prepare": "npm run build",
```
lumirlumir, line 8: I believe this change can also remove the `npm run build` step in the `release-please.yml` file.

https://github.com/eslint/js/blob/33a7d72c70ab93b9474635d03266649b49a5514a/.github/workflows/release-please.yml#L45

packages/eslint-scope/tsconfig.json
```diff
@@ -1,15 +1,9 @@
...
-    "noEmit": true,
-    "noImplicitThis": true,
-    "strictNullChecks": true,
-    "strictFunctionTypes": true,
-    "types": []
-  },
-  "include": ["lib"]
+    "noImplicitAny": false,
+    "rootDir": "lib",
+    "outDir": "dist"
```
lumirlumir:
```suggestion
    "outDir": "dist",
    "noEmit": true
```

Can we add `"noEmit": true` here? Currently, `eslint-scope` only uses `tsc` for type checking/type tests, not for emitting build output. However, running `test:types` in `eslint-scope` generates several `.d.ts` files under `dist`.

This only affects the dev environment right now, but if someone accidentally runs `test:types` before packaging or after a clean build, it could pollute the `dist` directory.

packages/eslint-scope/tests/types/tsconfig.json
```diff
@@ -1,8 +1,15 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "allowJs": false,
     "checkJs": false,
-    "strict": true
+    "noEmit": true,
```
lumirlumir:
```suggestion

```

Non-blocking suggestion: I think this could be removed if the previous comment is addressed.

.github/workflows/ci.yml
```diff
@@ -64,11 +60,42 @@ jobs:
...
+              run: npm install
+
+            - name: Test types (eslint-scope)
+              run: npm run test:types -w packages/eslint-scope
+
+            - name: Test types (eslint-visitor-keys)
+              run: npm run test:types -w packages/eslint-visitor-keys
+
+            - name: Test types (espree)
+              run: npm run test:types -w packages/espree
```
lumirlumir, line 88: I believe this check is redundant. All three packages' `test` scripts already include a `test:types` check, so this basic type check with the current TypeScript version is also covered by the `test` job above.

https://github.com/eslint/js/blob/33a7d72c70ab93b9474635d03266649b49a5514a/packages/eslint-scope/package.json#L42

https://github.com/eslint/js/blob/33a7d72c70ab93b9474635d03266649b49a5514a/packages/eslint-visitor-keys/package.json#L37

https://github.com/eslint/js/blob/33a7d72c70ab93b9474635d03266649b49a5514a/packages/espree/package.json#L64

---

One option would be to remove this check. If you prefer to keep this consistent with the setup in the [`rewrite`](https://github.com/eslint/rewrite/blob/main/.github/workflows/ci.yml#L81-L94) repository, another option would be to split the package scripts into `test:unit` and `test:types`, then run only `test:unit` in the `test` job and keep `test:types` in the `test_types` job, as done in https://github.com/eslint/rewrite/pull/357.

Reference: https://github.com/eslint/rewrite/blob/main/packages/config-array/package.json#L41-L45

lumirlumir (changes requested):

.github/workflows/ci.yml
sethamus: Went with splitting `test:unit` and `test:types` as suggested 👍 .

lumirlumir (dismissed):
LGTM, thanks for your patience and hard work on this PR.

Would like @fasttime to verify the changes before merging.

packages/eslint-scope/tests/types/tsconfig.json
```diff
@@ -1,8 +1,14 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "allowJs": false,
     "checkJs": false,
-    "strict": true
+    "rootDir": "../..",
+    "noImplicitAny": true,
+    "exactOptionalPropertyTypes": true,
```
fasttime: It looks like this is redundant now.

```suggestion
```

tsconfig.dist-legacy.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "allowJs": false,
+    "checkJs": false,
+    "emitDeclarationOnly": false,
+    "noEmit": true,
+    "strict": true,
+    "exactOptionalPropertyTypes": true,
```
fasttime: These settings should be also redundant:

```suggestion
```

tsconfig.dist.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "allowJs": false,
+    "checkJs": false,
+    "emitDeclarationOnly": false,
+    "noEmit": true,
+    "strict": true,
+    "exactOptionalPropertyTypes": true,
```
fasttime: Here also:

```suggestion
```

sethamus: Done, I removed the redundant settings.

lumirlumir (approved):
Re-approving.

fasttime (approved):
LGTM, thanks!

PR #747, opened 2026-05-07
ci: add Node.js 26 to CI

lumirlumir:
## Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

## AI acknowledgment

- [x] I did _not_ use AI to generate this PR.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

## What is the purpose of this pull request?

On May 5, 2026, Node.js 26 was released.

Ref: https://github.com/nodejs/node/releases/tag/v26.0.0

This PR is a straightforward CI matrix update to run tests on Node.js 26.

Just as dropping Node.js 25 from CI was suggested in https://github.com/eslint/eslintrc/pull/231#discussion_r3472949344, I’ve removed it as well.

## What changes did you make? (Give an overview)

I think `c8`’s `yargs` was not a blocker here. Even without adding the `test:coverage` script suggested in https://github.com/eslint/eslint/issues/20848, everything worked as expected, so I only updated the CI matrix and `mocha` to the latest.

Would it be better to add the `test:coverage` script, or is it fine to leave it as-is since CI is passing without that change?

## Related Issues

Related Issue: https://github.com/eslint/eslint/issues/20848

## Is there anything you'd like reviewers to focus on?

N/A

PR #757, opened 2026-06-28
test: separate coverage runs from default tests and replace `Makefile`

lumirlumir:
#### Prerequisites checklist

#### AI acknowledgment

- [x] I did _not_ use AI to generate this PR.
- [ ] (If the above is not checked) I have reviewed the AI-generated content before submitting.

This PR separates coverage runs from the default tests, as mentioned in https://github.com/eslint/js/pull/747#issuecomment-4816569042.

It mostly follows the same strategy used in https://github.com/eslint/rewrite/pull/357.

The current version of the PR mainly focuses on avoiding a large git diff. However, I have two questions:

- Should the `Makefile.js` for `eslint-scope` include the `mocha` test runner script? I think `mocha` could be defined directly in the `scripts` section of `package.json` instead of being consolidated into `Makefile.js`, since this is the only package that uses `Makefile.js`, unlike `eslint-visitor-keys` and `espree`.

- Should we use `--color --reporter progress` in `packages/espree/package.json`? The [language plugins](https://github.com/eslint/markdown/blob/main/package.json#L70) and [`rewrite`](https://github.com/eslint/rewrite/blob/main/packages/object-schema/package.json#L37) projects also don’t use this flag. I think we could consolidate `test:cjs` and `test:esm` into a `test:unit` script without this flag.

#### Related Issues

Ref: https://github.com/eslint/js/pull/747#issuecomment-4816569042

N/A


DMartens:
To your questions:
- I agree that removing the `Makefile` makes sense. As you said the test could be defined inline and the `checkLicenses` script could be moved into `tools/check-licenses.js`
- Both of the links do not show `mocha` using the mentioned CLI flags (maybe you modified them locally?). To the suggestion itself: I think having better defaults is worthwile but this should be uniform for all repositories

packages/eslint-visitor-keys/package.json
```diff
@@ -34,9 +34,10 @@
     "build:cjs": "rollup -c",
     "build:debug": "npm run build:cjs -- -m && npm run build:types",
     "build:types": "tsc -v && tsc",
-    "test": "mocha \"tests/**/*.test.cjs\" && mocha \"tests/**/*.test.js\" && npm run test:types",
-    "test:coverage": "c8 npm test",
-    "test:types": "tsd"
+    "test": "npm run test:types && npm run test:unit",
+    "test:coverage": "c8 npm run test:unit",
+    "test:types": "tsd",
+    "test:unit": "mocha \"tests/**/*.test.cjs\" && mocha \"tests/**/*.test.js\""
```
DMartens:
```suggestion
    "test:unit": "mocha \"tests/**/*.test.cjs\" \"tests/**/*.test.js\""
```
Is there a reason to split the run of tests?

eslint.config.js
```diff
@@ -40,18 +40,6 @@ export default defineConfig([
...
-	{
-		files: ["packages/eslint-scope/Makefile.js"],
-		languageOptions: {
-			globals: {
-				...globals.shelljs,
-				target: false,
-			},
-		},
-		rules: {
-			"no-console": "off",
```
lumirlumir, line 52: As `"no-console": "off"` is also specified in the following config object, I removed it entirely, since `shelljs` is no longer used in `eslint-scope`, as noted in https://github.com/eslint/js/pull/757#issuecomment-4904327453.

https://github.com/eslint/js/blob/db0b9e4c918ce8d77d25cc08d16274bd69426021/eslint.config.js#L84-L90

lumirlumir:
> I agree that removing the Makefile makes sense. As you said the test could be defined inline and the checkLicenses script could be moved into tools/check-licenses.js

I’ve updated it accordingly in https://github.com/eslint/js/pull/757/changes/6185414427c969e05cb14ba35a1d2c7eb92a0f49. I also removed the unnecessary `shelljs` dev dependency, since it was only used in a small part and can be replaced with native tools.

> Both of the links do not show mocha using the mentioned CLI flags (maybe you modified them locally?). To the suggestion itself: I think having better defaults is worthwile but this should be uniform for all repositories

Oops, I think my English was unclear. To clarify, I meant we should simply remove the `--color --reporter progress --timeout 30000` flags, since they aren’t generally used across the repositories.

I removed these flags in https://github.com/eslint/js/pull/757/changes/6185414427c969e05cb14ba35a1d2c7eb92a0f49, but please let me know if you have any other ideas.

For `--timeout 30000`, in [`markdown`](https://github.com/eslint/markdown/blob/main/package.json#L70), [it was kept due to an issue involving a super-linear regular expression and potential ReDoS concerns](https://github.com/eslint/markdown/pull/463#pullrequestreview-3023680643). However, in this repo, that timeout doesn’t seem necessary, so I removed it.

https://github.com/eslint/js/blob/db0b9e4c918ce8d77d25cc08d16274bd69426021/packages/espree/package.json#L66

packages/eslint-visitor-keys/package.json
lumirlumir: It was originally kept to reduce unnecessary diffs, but your suggestion makes sense, I’ve updated it as well in https://github.com/eslint/js/pull/757/commits/6185414427c969e05cb14ba35a1d2c7eb92a0f49.

packages/eslint-scope/tools/check-licenses.js
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return OPEN_SOURCE_LICENSES.some(license => license.test(licenses));
+}
+
+//------------------------------------------------------------------------------
+// License Checking
+//------------------------------------------------------------------------------
+
+console.log("Validating licenses");
```
DMartens: This `console.log` is now unnecessary for giving feedback as the script only checks the license (before it would test and check licenses)

DMartens (commented):
The removal of `shelljs` and extracting the license check script from the `Makefile.js` are big wins. As the PR no longer only updates the test scripts, we should update the PR title, e.g. "refactor: use package.json scripts over Makefile"

I do not have a strong opinion on setting defaults for the test runner such as `--color` or `--reporter` as I think the developer should choose for themselves how the test output should look. As such I agree with removing them, making it consistent with the other repositories.

Regarding the `timeout`: if the problem pops up again, we should rather use test suite / case scoped [timeouts](https://mochajs.org/features/timeouts/) rather than a global one (the global timeout may hide other problems)

packages/eslint-scope/tools/check-licenses.js
lumirlumir: Updated it in https://github.com/eslint/js/pull/757/changes/4fcaa90624f911055bc65081b4c1c3297097f0d5

lumirlumir:
Friendly ping @DMartens

DMartens (approved):
Changes LGTM, thanks. Sorry for the delay.

PR #765, opened 2026-07-21
docs: fix const reassignment in eslint-scope usage example

electrohyun:
#### Prerequisites checklist

#### AI acknowledgment

- [ ] I did _not_ use AI to generate this PR.
- [x] (If the above is not checked) I have reviewed the AI-generated content before submitting.

- Documentation update

Hello! This is a small fix for an example that currently runs into an error.

image

In the eslint-scope usage example, `currentScope` is declared with `const`, but it gets reassigned in the `estraverse.traverse` callbacks just below. Running the example as written throws `TypeError: Assignment to constant variable.`

image

Looking through the history, it seems this variable was originally declared with `var`, and it looks like it got turned into `const` when the example was converted during the ESM migration (eslint/eslint-scope#71). Since the variable is reassigned, `let` fits here.

Changed the `const` declaration to `let` (one line), so the example runs without the reassignment error.

#### Related Issues

None

Nothing in particular. It's a docs-only one-line change.

### DIFF
--- a/packages/eslint-scope/README.md
+++ b/packages/eslint-scope/README.md
@@ -56,7 +56,7 @@ const options = {
 const ast = espree.parse(code, { range: true, ...options });
 const scopeManager = eslintScope.analyze(ast, options);
 
-const currentScope = scopeManager.acquire(ast); // global scope
+let currentScope = scopeManager.acquire(ast); // global scope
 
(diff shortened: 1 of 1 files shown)


lumirlumir (approved):
Nice find, thanks!

